PR #18, opened 2020-07-06
Loops through a post request and gets a matching rights_shell to the request's identifier

### LINKED ISSUE #8: Loop through IDs from POST request, retrieve rights shells
For each Rights Shell ID in the POST request, retrieve the rights shell (including the calculation rules)

p-galligan:
Fixes #8 .

Loops through a post request through an API view and looks for each object's identifier. Gets the rights shell that matches that identifier exactly (I'm assuming there should only be one match per identifier/shell), and then appends the shell to a list.

Prints out an exception right now if it cannot match identifiers.

Also includes a first stab at testing one of the models/views.

### DIFF
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -13,11 +13,10 @@ Including another URLconf
     1. Import the include() function: from django.urls import include, path
     2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
 """
+from assign_rights.views import RightsAssemblerView
 from django.contrib import admin
 from django.urls import include, path
 from rest_framework import routers
-
-from assign_rights.views import RightsAssemblerView
 
 router = routers.DefaultRouter()
 
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -14,7 +14,7 @@ Including another URLconf
     2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
 """
 from django.contrib import admin
-from django.urls import path
+from django.urls import include, path
 from rest_framework import routers
 
 from assign_rights.views import RightsAssemblerView
@@ -24,5 +24,5 @@ router = routers.DefaultRouter()
 urlpatterns = [
     path('admin/', admin.site.urls),
     path('', include(router.urls)),
-    path('api/rights-assemble/', RightsAssembler.as_view(), name='rights-assemble')
+    path('api/rights-assemble/', RightsAssemblerView.as_view(), name='rights-assemble')
 ]
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -16,12 +16,8 @@ Including another URLconf
 from assign_rights.views import RightsAssemblerView
 from django.contrib import admin
 from django.urls import include, path
-from rest_framework import routers
-
-router = routers.DefaultRouter()
 
 urlpatterns = [
     path('admin/', admin.site.urls),
-    path('', include(router.urls)),
     path('api/rights-assemble/', RightsAssemblerView.as_view(), name='rights-assemble')
 ]
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -15,7 +15,14 @@ Including another URLconf
 """
 from django.contrib import admin
 from django.urls import path
+from rest_framework import routers
+
+from assign_rights.views import RightsAssemblerView
+
+router = routers.DefaultRouter()
 
 urlpatterns = [
     path('admin/', admin.site.urls),
+    path('', include(router.urls)),
+    path('api/rights-assemble/', RightsAssembler.as_view(), name='rights-assemble')
 ]
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -15,7 +15,7 @@ Including another URLconf
 """
 from assign_rights.views import RightsAssemblerView
 from django.contrib import admin
-from django.urls import include, path
+from django.urls import path
 
 urlpatterns = [
     path('admin/', admin.site.urls),
--- a/assign_rights/assemble.py
+++ b/assign_rights/assemble.py
@@ -8,10 +8,12 @@ from .models import RightsShell
 class RightsAssembler(object):
     """docstring for RightsCalculator"""
 
-    def retrieve_rights(self, item):
+    def retrieve_rights(self, item, shell_list):
         """docstring for retrieve_rights"""
-        shell = RightsShell.objects.get(rights_id=id)
-        return shell
+        identifier = item.get("identifier")
+        shell = RightsShell.objects.get(rights_id=identifier)
+        shell_list.append(shell)
+        return shell_list
 
     def calculate_dates(self):
         """docstring for calculate_dates"""
@@ -26,9 +28,10 @@ class RightsAssembler(object):
     pass
 
     def run(self, request_list):
+        shell_list = []
         for item in request_list:
             try:
-                self.retrieve_rights(item)
-                return 'test'
+                self.retrieve_rights(item, shell_list)
             except Exception as e:
-                print("Could not find matching Rights Shell")
+                print("Error retrieving rights shell: {}".format(str(e)))
+        print(shell_list)
--- a/assign_rights/assemble.py
+++ b/assign_rights/assemble.py
@@ -12,7 +12,7 @@ class RightsAssembler(object):
         """Retrieves a rights shell whose rights_id matches an identifier from
         a post request.
         """
-        return RightsShell.objects.get(rights_id=identifier)
+        return RightsShell.objects.get(pk=identifier)
 
     def calculate_dates(self):
         """docstring for calculate_dates"""
--- a/assign_rights/assemble.py
+++ b/assign_rights/assemble.py
@@ -14,11 +14,7 @@ class RightsAssembler(object):
         """
         rights_shells = []
         for ident in rights_ids:
-            try:
-                rights_shells.append(RightsShell.objects.get(pk=ident))
-            except RightsShell.DoesNotExist:
-                error = "Could not find matching shell with identifier: {}".format(str(ident))
-                print(error)
+            rights_shells.append(RightsShell.objects.get(pk=ident))
         return rights_shells
 
     def calculate_dates(self, end_date):
@@ -38,5 +34,5 @@ class RightsAssembler(object):
             rights_shells = self.retrieve_rights(rights_ids)
             for shell in rights_shells:
                 self.calculate_dates(end_date)
-        except Exception as e:
+        except RightsShell.DoesNotExist as e:
             print("Error retrieving rights shell: {}".format(str(e)))
--- a/assign_rights/assemble.py
+++ b/assign_rights/assemble.py
@@ -1,3 +1,4 @@
+from django.core.exceptions import ObjectDoesNotExist
 from .models import RightsShell
 
 # Receive POST request that contains multiple rights IDs and one date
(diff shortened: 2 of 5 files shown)


aquila/urls.py
```diff
@@ -13,9 +13,15 @@
     1. Import the include() function: from django.urls import include, path
     2. Add a URL to urlpatterns:  path('blog/', include('blog.urls'))
 """
+from assign_rights.views import RightsAssemblerView
 from django.contrib import admin
-from django.urls import path
+from django.urls import include, path
+from rest_framework import routers
+
+router = routers.DefaultRouter()
```
helrond: Do we need this? We're not using ModelViewSets as far as I know, which is the only reason you'd need this...

aquila/urls.py
```diff
@@ -13,9 +13,15 @@
...
 
 urlpatterns = [
     path('admin/', admin.site.urls),
+    path('', include(router.urls)),
```
helrond: See above - may be unnecessary.

assign_rights/tests.py
```diff
@@ -1,3 +1,38 @@
...
+        s = self.create_rightsshell()
+        self.assertTrue(isinstance(s, RightsShell))
+
+
+class TestViews(TestCase):
+
+    def test_rightsassemblerview(self):
+        TestModels.create_rightsshell(self)
+        factory = APIRequestFactory()
+        request = factory.post(reverse('rights-assemble'), {"objects": [{"identifier": 1, "start_date": "2019-12-09", "end_date": "2020-01-01"}, {"identifier": 2, "start_date": "2019-12-01", "end_date": "2020-02-01"}, {"identifier": 3, "start_date": "2019-11-09", "end_date": "2020-03-01"}]}, format='json')
```
helrond: @bonniegee can you comment on this data structure? My (untested) assumption was that the data would look more like:
```
{"identifiers": [1, 2, 3, 4], "start_date": "2019-01-01", "end_date": "2019-06-01"}
```

assign_rights/assemble.py
```diff
@@ -6,9 +8,12 @@
 class RightsAssembler(object):
     """docstring for RightsCalculator"""
 
-    def retrieve_rights(self):
+    def retrieve_rights(self, item, shell_list):
```
helrond:
```suggestion
    def retrieve_rights(self, item):
```
I'd append to `shell_list` in the `for` loop below rather than passing it in here.

assign_rights/assemble.py
```diff
@@ -6,9 +8,12 @@
...
         """docstring for retrieve_rights"""
-    pass
+        identifier = item.get("identifier")
```
helrond: See my comment on the tests data below - `item` _may_ in this case be an identifier.

assign_rights/assemble.py
```diff
@@ -6,9 +8,12 @@
...
+        shell = RightsShell.objects.get(rights_id=identifier)
+        shell_list.append(shell)
+        return shell_list
```
helrond:
```suggestion
        return RightsShell.objects.get(rights_id=identifier)
```
If you follow my suggestion above, I think you can do this in one line. 

Also, @bonniegee can confirm, but you may want to simply target the model `pk`, because I think `rights_id` may be going away as a separate field.

assign_rights/assemble.py
```diff
@@ -6,9 +8,12 @@
 class RightsAssembler(object):
     """docstring for RightsCalculator"""
 
-    def retrieve_rights(self):
+    def retrieve_rights(self, item, shell_list):
         """docstring for retrieve_rights"""
```
helrond: Docstring plz

assign_rights/assemble.py
```diff
@@ -21,3 +26,12 @@ def create_json(self):
     def return_rights(self):
         """docstring for return_rights"""
     pass
+
+    def run(self, request_list):
+        shell_list = []
+        for item in request_list:
+            try:
+                self.retrieve_rights(item, shell_list)
```
helrond:
```suggestion
                shell_list.append(self.retrieve_rights(item))
```

assign_rights/assemble.py
```diff
@@ -21,3 +26,12 @@ def create_json(self):
...
+            except Exception as e:
```
helrond: As this method continues to get built out, we'll want to target specific exceptions, for example an id that doesn't match any RightsShell objects will raise a `RightsShell.NotFoundError` (I think). But that can wait until later.

assign_rights/tests.py
```diff
@@ -1,3 +1,38 @@
...
+            start_date_period=None,
+            end_date_period=None,
+            end_date_open=False,
+            license_terms=None,
+            statute_citation=None
+        )
+
+    def test_rightsshell_creation(self):
+        s = self.create_rightsshell()
+        self.assertTrue(isinstance(s, RightsShell))
```
helrond: I'm not convinced this test is doing all that much - we're basically just testing that Django's ORM works, which is something that's done much more effectively elsewhere.

assign_rights/tests.py
```diff
@@ -1,3 +1,38 @@
...
+
+
+class TestViews(TestCase):
+
+    def test_rightsassemblerview(self):
+        TestModels.create_rightsshell(self)
```
helrond: Instead of calling another test class, I'd suggest adding a `setUp` method to this class which creates the rights statements desired.

assign_rights/views.py
```diff
@@ -1,42 +1,57 @@
...
+
+class RightsAssemblerView(APIView):
+    '''
+    Calls the RightsAssembler class from assemblers.
+    '''
+
+    def post(self, request, format=None):
+        request_list = request.data.get('objects')
+        shell_list = RightsAssembler().run(request_list)
+        return Response(request_list, status=200)
```
helrond: What happens when there's an error? Shouldn't your view return something other than a 200?

helrond (changes requested):
OK, we're on our way! I think there are a bunch of places here where things can be simplified. LMK if you have questions!

assign_rights/assemble.py
p-galligan: Right now if there is an error that gets sent: `Error retrieving rights shell: RightsShell matching query does not exist.`

assign_rights/tests.py
bonniegee: You're right that there would be an array of identifiers (I'm realizing I implied otherwise @p-galligan, my bad!). However, there would only be an end date, since a start date wouldn't be needed in order to calculate rights statement dates.

assign_rights/assemble.py
helrond: Right, but we might want to return a 404 if we can't find a rights shell, or a 500 if something else goes wrong. Exception handling is something that is usually built out incrementally though, so it's more something to keep in mind and less something to act on right now.

assign_rights/tests.py
```diff
@@ -1,3 +1,38 @@
...
+            license_terms=None,
+            statute_citation=None
+        )
+
+    def test_rightsshell_creation(self):
+        s = self.create_rightsshell()
+        self.assertTrue(isinstance(s, RightsShell))
+
+
+class TestViews(TestCase):
```
helrond: I don't think this test asserts anything - so it's executing the code but you're not actually testing the output of anything.

A couple of possibilities:
1. the response code from the request
2. something in the response data (like an expected string or number of things)

assign_rights/views.py
```diff
@@ -1,42 +1,56 @@
...
+
+class RightsAssemblerView(APIView):
+    '''
+    Calls the RightsAssembler class from assemblers.
+    '''
+
+    def post(self, request, format=None):
+        request_list = request.data
+        shell_list = RightsAssembler().run(request_list)
+        return Response(request_list, status=200)
```
helrond:
```suggestion
        return Response(shell_list, status=200)
```

assign_rights/views.py
```diff
@@ -1,42 +1,56 @@
...
     pass
+
+
+class RightsAssemblerView(APIView):
+    '''
+    Calls the RightsAssembler class from assemblers.
+    '''
+
+    def post(self, request, format=None):
+        request_list = request.data
```
helrond: I would suggest parsing the request data here, so 
```suggestion
        rights_ids = request.data.get("identifiers")
        start_date = request.data.get("start_date")
        end_date = request.data.get("end_date")
        assembled = RightsAssembler().run(rights_ids, start_date, end_date)
```

assign_rights/assemble.py
bonniegee: @helrond is correct, `rights_id` will be removed and we'll use `pk` to refer to rights shells

p-galligan:
@helrond requesting re-review. I think the exceptions/tests could probably use some work, but I'm not really sure how to improve them at my skill level.

assign_rights/views.py
```diff
@@ -1,42 +1,57 @@
-from django.shortcuts import render
 from django.views.generic import CreateView, DetailView, ListView, UpdateView
+from rest_framework.response import Response
+from rest_framework.views import APIView
 
+from .assemble import RightsAssembler
 
-def RightsShellsListView(ListView):
+
+class RightsShellsListView(ListView):
```
bonniegee: This was my error, but if these views are (correctly) being updated from methods to classes, they should also change from plural to singular

bonniegee (changes requested):
This looks good! The only changes I would suggest are a result of my earlier error

assign_rights/views.py
p-galligan: so `class RightsShellListView` and `GroupingListView` instead of `GroupingsListView`?

p-galligan: Nevermind, just checked your branch. Will update to match those.

bonniegee (approved):

assign_rights/assemble.py
```diff
@@ -6,9 +8,11 @@
...
     """docstring for RightsCalculator"""
 
-    def retrieve_rights(self):
-        """docstring for retrieve_rights"""
-    pass
+    def retrieve_rights(self, identifier):
+        """Retrieves a rights shell whose rights_id matches an identifier from
+        a post request.
+        """
+        return RightsShell.objects.get(pk=identifier)
```
helrond: After refactoring, this is now a one-line function, so I think there are two possible paths here:
1. move the line to the `run` method.
2. move the `for` loop currently in the `run` method into this function.

Thinking about it, my preference would be for the latter, because you need to know you have all the rights shells before processing, so it's more efficient if you make sure those all exist first before trying to calculate dates and/or assemble the statement.

So this function would be:
```suggestion
    def retrieve_rights(self, rights_ids):
        """Retrieves a rights shell whose rights_id matches an identifier from
        a post request.
        """
        rights_shells = []
        for ident in rights_ids:
            rights_shells.append(RightsShell.objects.get(pk=ident))
        return rights_shells
```

assign_rights/assemble.py
```diff
@@ -21,3 +25,12 @@ def create_json(self):
     def return_rights(self):
         """docstring for return_rights"""
     pass
+
+    def run(self, rights_ids, end_date):
+        shells = []
+        for identifier in rights_ids:
+            try:
+                shells.append(self.retrieve_rights(identifier))
```
helrond: Based on my comment above, if you opt to go that route, this function would then need to be amended as well:
```suggestion

    def run(self, rights_ids, end_date):
            try:
                rights_shells = self.retrieve_rights(rights_ids))
                for shell in rights_shells:
                    ...
```

helrond (commented):
There's an option to change the logic a bit here if you want. Take a look.

In terms of tests, I think we should actually be testing the RightsAssembler class, rather than the view. I can add some commits to this branch to demonstrate that.

assign_rights/assemble.py
p-galligan: Ok, latter option makes more sense to me as well. I just spent 15 minutes thinking about the best way to calculate dates with the current for loop and it wasn't going to be great.

p-galligan:
> There's an option to change the logic a bit here if you want. Take a look.
> 
> In terms of tests, I think we should actually be testing the RightsAssembler class, rather than the view. I can add some commits to this branch to demonstrate that.

@helrond Sure, I wouldn't mind an example of testing the `retrieve_rights` class!

helrond:
OK, I added a new TestCase for AssembleRights. Right now it just tests the `retrieve_rights` method but in the future it could test other methods. I also removed the exception handling from `retrieve_rights` since we want this to throw an exception if it can't find all of the rights shells so we can handle that in the `run` method.

helrond (approved):

p-galligan:
Thank you! These tests make sense, and I wasn't aware of auto-generating random values in the setUp, that's super helpful.

HaSistrunk (approved):

PR #13, opened 2020-06-30
Add login and logout views, plus one authenticated view

helrond:
In order to test this:

1. Bring the app up (making sure migrations run)
2. Navigate to http://localhost:8000/logged-in and observe that you are redirected to http://localhost:8000/login. This is a result of the `LoginRequiredMixin` applied to the `LoggedInView`.
3. In a separate terminal, open up the Django shell and create a user.
```
docker-compose exec aquila-web python manage.py shell
from assign_rights.models import User
User.objects.create_user("test", "test@example.com", "testpassword")
```
4. Navigate to http://localhost:8000/login and enter the username and password from the user you created above. You should be redirected to http://localhost:8000/logged-in
5. Using the link on that page, log out. You should be redirected to http://localhost:8000/login.

A couple of notes on this:
- I added a `base.html` template which provides a slot for a content block, but will need to be updated.
- The `LoggedInView` is for purposes of demonstration only, and will need to be replaced with a dashboard or some other view.
- I've added tests in a `TestAssignRightsViews` class.

fixes #11

PR #19, opened 2020-07-06
Builds out base template, adds page title mixin

### LINKED ISSUE #14: Build out base.html template
The `base.html` template currently only provides slots for a content block, but ideally it would also provide a place for additional blocks, including head content, navigation, scripts, etc. It would also be good to review this template with an eye to ensuring that HTML structure is sound.

In order to move forward on this, a quick wireframe of common page elements (nav, header, footer, etc) would be useful.

helrond:
Builds out `base.html` with additional includes for head, navbar and scripts.
Adds Bootstrap files (CSS and JS) to static directory
Adds a PageTitleMixin to simplify process of adding page titles to views (used in templates in multiple places)

**NOTE:** This is branched from #13, so that should be merged first.

fixes #14


bonniegee (approved):

assign_rights/templates/base.html
```diff
@@ -0,0 +1,35 @@
...
+    {% block extra_css %}{% endblock %}
+  </head>
+  <body>
+    <div class="wrapper">
+
+      {% include 'navbar.html' %}
+
+      <div class="content-wrapper container">
+
+        <!-- Content Header (Page header) -->
```
HaSistrunk: I think it makes sense to move the h1 into `<main>` wrapped in `<div>` instead of `<section>` elements. Edit: sorry, this is selecting the code in a weird place, but hopefully this makes sense.

assign_rights/templates/navbar.html
```diff
@@ -0,0 +1,29 @@
+{% load i18n %}
+
+<header class="main-header">
+
+  <nav class="navbar navbar-expand-sm navbar-dark bg-dark mb-2" role="navigation" aria-label="{% trans 'navbar' %}">
```
HaSistrunk:
```suggestion
  <nav class="navbar navbar-expand-sm navbar-dark bg-dark mb-2" role="navigation">
```

assign_rights/templates/users/login.html
```diff
@@ -0,0 +1,25 @@
...
+<div class="row justify-content-sm-center">
+  <div class="card col-sm-6 p-4">
+    <form action="." method="post">
+      {{ form.non_field_errors }}
+
+      {% csrf_token %}
+
+      <div class="form-group has-feedback">
+        {{ form.username.errors }}
+        <input type="text" name="username" id="{{ form.username.id_for_label }}" class="form-control" placeholder="Username" value="{% if form.username.value %}{{ form.username.value }}{% endif %}" />
```
HaSistrunk: Both inputs in the form need associated form labels where the "for" value matches the input id: `<label for="id_for_label">Username</label>`. The placeholder text is not necessary/not optimal. See https://webaim.org/techniques/forms/controls

HaSistrunk (changes requested):

HaSistrunk (approved):

p-galligan (approved):

PR #23, opened 2020-07-08
Adds Grouping Views

### LINKED ISSUE #6: Add Grouping Browse Page
In order to see whether a grouping already exists and retrieve associated rights shell IDs, a user should be able to browse and search existing groupings. This should eventually be searchable, sortable, and filterable, but should initially successfully display data. The following data should be displayed in the grouping browse:

* Title
* Description
* Associated Rights IDs

This involves creating a view that returns a list of objects from the database and then create a template that writes that list to an HTML table.

### LINKED ISSUE #7: UI for adding a grouping
One major piece of functionality for this application is creating groupings, which can then be associated with rights shells. his form should include all fields in the Grouping model.

### LINKED ISSUE #15: Replaced LoggedInView with a dashboard view
Currently `LoggedInView` exists only to demonstrate the flow of users logging into the app. However, after logging in, users probably expect to go to some sort of dashboard page. 

A quick wireframe of what this page should look like would be helpful.

helrond:
Adds views and templates for Groupings, including:
- A list of all Grouping objects, fixes #6 
- A single Grouping object
- Forms to create and update Grouping objects, fixes #7 

Redirects users to GroupingsListView on login, fixes #15


bonniegee (approved):

p-galligan (approved):

HaSistrunk:
How long are we thinking these descriptions might be when creating groupings? I'm wondering if that should be a `<textarea>` instead of `<input>` like (I think) is in the wireframe.

HaSistrunk:
So this Rights Shell select is a placeholder right now, yeah?  
Screen Shot 2020-07-08 at 2 57 53 PM

helrond:
> How long are we thinking these descriptions might when creating groupings? I'm wondering if that should be a `<textarea>` instead of `<input>` like (I think) is in the wireframe.

The default widget in Django is based on the field in the model; in this case it's a `CharField` with a max of 255 characters. It's possible to specify particular widgets in the forms. I was thinking we should hold off on that work for now though because it's likely that we'll want to look at those issue across both Groupings and RightsShells.

> So this Rights Shell select is a placeholder right now, yeah?

Sort of - I think your instance might not have any RightsShells, which is why you're seeing that weird line. Again, I'd like to look at these display issues more holistically, perhaps with a library like [django-crispy-forms](https://django-crispy-forms.readthedocs.io/en/latest/)

HaSistrunk (approved):

PR #24, opened 2020-07-08
WIP: Add UI for browsing, creating, updating, and viewing rights shells

bonniegee:
Adds very basic pages to create, update, view, and browse rights shells. Among other work, needs to include rights granted fields.

@helrond , I'm having trouble figuring out how to use formsets in my views. Do I need to create a formset for rights shells in addition to rights granted? I got as far as updating `forms.py`, but wasn't sure what to add to `views.py`.


helrond:
@bonniegee I think you need to use [inline formsets](https://docs.djangoproject.com/en/3.0/topics/forms/modelforms/#inline-formsets).

PR #26, opened 2020-07-09
Calculate dates and return rights dates for shells and acts

p-galligan:
Sets up logic for calculating end dates for RightsShells and associated RightsGrants.

@helrond @bonniegee could use another set of eyes on this to see if the calculation logic makes sense here.


assign_rights/assemble.py
```diff
@@ -17,9 +19,38 @@ def retrieve_rights(self, rights_ids):
             rights_shells.append(RightsShell.objects.get(pk=ident))
         return rights_shells
 
-    def calculate_dates(self, end_date):
+    def calculate_dates(self, shell, end_date):
         """docstring for calculate_dates"""
```
helrond: A docstring indicating what you intend this method to do would be helpful in terms of evaluating if it's actually doing that...

assign_rights/assemble.py
```diff
@@ -17,9 +19,38 @@ def retrieve_rights(self, rights_ids):
...
+            if not grant.end_date_open:
+                period = grant.end_date_period
+                if not grant.end_date:
+                    act["grant_id"] = grant.pk
+                    act["act_end"] = end_date + relativedelta(years=period)
+                    grants_end.append(act)
+                else:
+                    act["grant_id"] = grant.pk
+                    act["act_end"] = grant.end_date + relativedelta(years=period)
+                    grants_end.append(act)
```
helrond: I think you could avoid the duplicated logic in each of these branches by setting a variable for the end date that's used in calculation.

assign_rights/assemble.py
```diff
@@ -17,9 +19,38 @@ def retrieve_rights(self, rights_ids):
             rights_shells.append(RightsShell.objects.get(pk=ident))
         return rights_shells
 
-    def calculate_dates(self, end_date):
+    def calculate_dates(self, shell, end_date):
```
helrond: Would suggest naming the `end_date` variable to something that distinguishes it a bit from the other end dates in this method, like maybe `grouping_end_date`

assign_rights/assemble.py
```diff
@@ -17,9 +19,38 @@ def retrieve_rights(self, rights_ids):
...
+                    act["grant_id"] = grant.pk
+                    act["act_end"] = grant.end_date + relativedelta(years=period)
+                    grants_end.append(act)
+            else:
+                act["grant_id"] = grant.pk
+                act["act_end"] = None
+        rights_info["shell_end"] = shell_end
+        rights_info["acts"] = grants_end
+        print(rights_info)
+        return rights_info
```
helrond: So it seems like this is returning a fairly complex dict. I wonder if it would make sense to take a more atomic approach and call the `calculate_dates` and `create_json` methods for each RightsShell AS WELL AS each RightsGranted associated with that shell? If you did that (and with some changes to model fields) you could substantially trim down the logic here, and I think the resulting inputs and outputs would be clearer and easier to test/maintain.

assign_rights/models.py
```diff
@@ -26,9 +26,11 @@ class RightsShell(models.Model):
     applicable_start_date = models.DateField(blank=True, null=True)
     applicable_end_date = models.DateField(blank=True, null=True)
     start_date_period = models.PositiveSmallIntegerField(
-        blank=True, null=True
+        blank=True, null=True, default=0
+    )
+    end_date_period = models.PositiveSmallIntegerField(
+        blank=True, null=True, default=0
```
helrond: If you're setting a default you should remove the blank/null attributes

assign_rights/models.py
```diff
@@ -56,8 +58,12 @@ class RightsGranted(models.Model):
...
     start_date = models.DateField(blank=True, null=True)
     end_date = models.DateField(blank=True, null=True)
-    start_date_period = models.PositiveSmallIntegerField(blank=True, null=True)
-    end_date_period = models.PositiveSmallIntegerField(blank=True, null=True)
+    start_date_period = models.PositiveSmallIntegerField(
+        blank=True, null=True, default=0
+    )
+    end_date_period = models.PositiveSmallIntegerField(
+        blank=True, null=True, default=0
+    )
```
helrond: See above

assign_rights/tests.py
```diff
@@ -44,14 +44,28 @@ def add_rights_shells(count=5):
             determination_date=random_date(),
             note=random_string(),
             start_date=random_date(),
-            end_date=random_date(),
+            end_date=random.choice([None, random_date()]),
             start_date_period=None,
             end_date_period=random.randint(0, 10),
-            end_date_open=False,
+            end_date_open=random.choice(["True", "False"]),
```
helrond: Should these really be strings?

assign_rights/tests.py
```diff
@@ -44,14 +44,28 @@ def add_rights_shells(count=5):
...
+    for x in range(count):
+        RightsGranted.objects.create(
+            basis=random.choice(RightsShell.objects.all()),
+            act=random.choice(["publish", "disseminate", "replicate", "migrate", "modify", "use", "delete"]),
+            restriction=random.choice(["allow", "disallow", "conditional"]),
+            start_date=random_date(),
+            end_date=random.choice([None, random_date()]),
+            start_date_period=None,
+            end_date_period=random.randint(0, 10),
+            end_date_open=random.choice(["True", "False"]),
```
helrond: Ditto from above - should these be strings or actual booleans?

assign_rights/tests.py
p-galligan: Nope. Thanks for catching that. Just pushed up a fix.

assign_rights/tests.py
```diff
@@ -95,21 +109,60 @@ def test_grouping_form(self):
...
+
+    def test_calculate_dates(self):
+        """Tests the calculate_dates method.
+
+        Asserts the method returns a DateTimeField or None if the end date is open.
+        Asserts that the relative delta of the calculated date and the correct end date
+        is equal to the end date period of the object.
+        """
+        assembled = self.assembler.retrieve_rights(self.rights_ids)
+        # grouping_end_date = random_date() wasn't working here because it returned a string
```
helrond: `random_date` could easily be updated so it returns a datetime object...

assign_rights/tests.py
```diff
@@ -95,21 +109,60 @@ def test_grouping_form(self):
...
+            assembled = self.assembler.retrieve_rights(self.rights_ids)
+
+    def test_calculate_dates(self):
+        """Tests the calculate_dates method.
+
+        Asserts the method returns a DateTimeField or None if the end date is open.
+        Asserts that the relative delta of the calculated date and the correct end date
+        is equal to the end date period of the object.
+        """
+        assembled = self.assembler.retrieve_rights(self.rights_ids)
```
helrond: I would suggest just using a Django queryset `RightsShell.objects.all()` that way you keep your tests isolated and a failure in `test_retrieve_rights` won't cause a failure in `test_calculate_dates`.

assign_rights/tests.py
```diff
@@ -95,21 +109,60 @@ def test_grouping_form(self):
...
+        for object in assembled:
+            object_date = self.assembler.calculate_dates(object, grouping_end_date)
+            if object.end_date_open:
+                self.assertEqual(object_date, None)
+            elif object.end_date:
+                self.assertEqual(relativedelta(object_date, object.end_date).years, object.end_date_period)
+                self.assertTrue(isinstance(object_date, date))
+            else:
+                self.assertEqual(relativedelta(object_date, grouping_end_date).years, object.end_date_period)
+                self.assertTrue(isinstance(object_date, date))
```
helrond: These lines and 156-165 below look to be identical, so I'd suggest factoring them out to a separate function that gets called twice.

assign_rights/tests.py
p-galligan: Removed the `.isoformat` from `random_date` since it was in ISO format already, and all this was doing was transforming it into a string.

helrond (approved):
This is looking ✨ ! Thanks for sticking with it - all that refactoring really makes things much easier to understand.

bonniegee:
This looks good--I agree with Hillel that it's easy to understand. Are you planning on addressing start dates in a separate PR?

p-galligan:
@bonniegee I was working on only end dates based off of this comment: https://github.com/RockefellerArchiveCenter/aquila/pull/18#discussion_r450511774

I think start dates should be fairly easy to calculate, but I think we would also want to pass more information in through the POST request. Adding start date calculations would require reworking some tests and the type of data we're returning.

@helrond , we briefly talked about how we might need more info for start date calculations. Do you have thoughts on what that would be?

Also, I realized I forgot to change the `end_date_period` and `start_date_period` default values to 0, so I'm going to make a migration for that.

helrond:
@p-galligan I think start date calculations would essentially work the same way as end date calculations. But that means that Zorya would need to send a start date for the groupings, so it could be passed into the appropriate method(s).

assign_rights/assemble.py
```diff
@@ -6,7 +8,11 @@
 
 
 class RightsAssembler(object):
-    """docstring for RightsCalculator"""
+    """Reads through a list of ids from a POST request and gets matching RightsShells
+    for each identifier (matching against the primary key). Gets all RightsGranted
+    objects related to each shell. Then, calculates rights end dates based on on info
```
HaSistrunk:
```suggestion
    objects related to each shell. Then, calculates rights end dates based on info
```

HaSistrunk (approved):

assign_rights/tests.py
```diff
@@ -95,21 +109,51 @@ def test_grouping_form(self):
...
 
-        rights_ids.append(len(rights_ids) + 1)
+        deleted = random.choice(RightsShell.objects.all())
+        deleted.delete()
+        self.rights_ids.append(deleted.pk)
         with self.assertRaises(RightsShell.DoesNotExist):
-            assembled = self.assembler.retrieve_rights(rights_ids)
+            assembled = self.assembler.retrieve_rights(self.rights_ids)
+
+    def check_end_dates(self, assembled, grouping_end_date):
```
helrond: I've been thinking about this and I think that instead of randomly setting attributes on the object, you may want to explicitly set those and then check for expected outputs. So, for example, set `end_date_open` to True and then assert that `calculate_dates` returns `None`.

That way you can be sure that you're consistently testing all the branches of the logic, and it's also a little clearer for someone coming to it without prior knowledge.

bonniegee:
The start date confusion was my error; Zorya will be sending start dates.

assign_rights/assemble.py
```diff
@@ -17,9 +23,21 @@ def retrieve_rights(self, rights_ids):
             rights_shells.append(RightsShell.objects.get(pk=ident))
         return rights_shells
 
-    def calculate_dates(self, end_date):
-        """docstring for calculate_dates"""
-    pass
+    def calculate_dates(self, object, grouping_end_date):
```
bonniegee: "Grouping" here refers to the content described in the POST request; I'm wondering if this variable name should be changed to avoid confusion.

assign_rights/tests.py
p-galligan: Ok, I've pushed up my latest changes. I'm not super happy with the tests right now, but I'm a little confused about how we'd go about doing the above while still keeping automatic generation. With start dates added, there are somewhere like 6 different start date and end date permutations for each RightsShells and RightsGranted.

assign_rights/assemble.py
```diff
@@ -17,9 +23,27 @@ def retrieve_rights(self, rights_ids):
...
+        date is open, and then based on the period and end_dates in the object.
+
+        If no end date in the object, use the grouping end date.
+        """
+        start_period = object.start_date_period
+        end_period = object.end_date_period
+        if not object.start_date:
+            object_start = request_start_date + relativedelta(years=start_period)
+        if object.start_date:
+            object_start = object.start_date + relativedelta(years=start_period)
```
helrond: So again you've got an opportunity to factor out common logic, here and lines 40-43.

assign_rights/assemble.py
```diff
@@ -17,9 +23,27 @@ def retrieve_rights(self, rights_ids):
             rights_shells.append(RightsShell.objects.get(pk=ident))
         return rights_shells
 
-    def calculate_dates(self, end_date):
-        """docstring for calculate_dates"""
-    pass
+    def calculate_dates(self, object, request_start_date, request_end_date):
+        """Calculate rights end dates for a given object based on whether the end
```
helrond: Docstring needs updating

assign_rights/assemble.py
```diff
@@ -29,10 +53,15 @@ def return_rights(self):
...
     pass
 
-    def run(self, rights_ids, end_date):
+    def run(self, rights_ids, request_start_date, request_end_date):
         try:
             rights_shells = self.retrieve_rights(rights_ids)
             for shell in rights_shells:
-                self.calculate_dates(end_date)
+                act_dates = []
+                self.calculate_dates(shell, request_end_date)
```
helrond: You'll need to pass `request_start_date` in here (and in line 64 below)

assign_rights/tests.py
helrond: Why don't you get the rest of this settled and then I can work with you on tests.

assign_rights/assemble.py
p-galligan: @helrond I'm not sure if my changes are what you were going for, but I mixed it up so that we just automatically use the request dates and then change them based on some leaner logic.

helrond: @p-galligan What I was saying was that the logic in lines 34-37 and 40-43 is exactly the same, the only difference is which attribute is being targeted. I think you can factor that out into a separate function, and then just call it twice with the relevant arguments.

Also, it looks like you're defining `end_period` twice, once in line 33 and then line 39.

p-galligan: Ok, I think I've worked out the easiest way to make this work. Let me know your thoughts.

p-galligan:
@helrond I had missed that you reworked the tests in here too until this morning. So you've moved some things to test_helpers and then are setting specific fields and rerunning some tests based on that.

I feel like this is ready for re-review/merging. Will request.

helrond (commented):
OK, I just made some changes to the docstrings and in doing that, noticed an issue with what's passed in to `get_date_value` - right now the tests assume that is a `date` object, rather than a string representation. HOWEVER, that data is going to be coming in as a string via a request, and it is nowhere (yet) converted to a `date` object. So I think some consideration of where that date gets converted needs to happen, plus updating tests and docstrings as necessary....

p-galligan:
That makes a lot of sense to me. I can mock those up quickly after work. My instinct is to make that conversion in the assemblers right before those dates would be used by something else. Thoughts on that?

helrond:
> That makes a lot of sense to me. I can mock those up quickly after work. My instinct is to make that conversion in the assemblers right before those dates would be used by something else. Thoughts on that?

If you do it in `get_date_value` you only have to do it once. Again, you'd have to change docstrings and tests in that case.

p-galligan:
Ok, so I got these updated, and I updated a couple of the tests. I felt this was lower-impact than updating the whole `random_date` method and then converting _from_ str to datetime in more instances, but I'm open to other ideas. Basically just turns two variables to strings in the tests and then turns them into date objects used in `AssertEqual` tests.

bonniegee (approved):

assign_rights/assemble.py
```diff
@@ -20,11 +21,14 @@ def get_date_value(self, object, field_name, request_date, period):
         Args:
             object: the object for which a date is to be calculated.
             field_name: the object attribute containing date data.
-            request_date: date data sumbitted in a request.
+            request_date: representation of a date sumbitted in a request.
             period: the numbe of years to be used in calculating the date.
```
helrond: Let's add the types of these args in parentheses. Like `object (obj)`

assign_rights/assemble.py
```diff
@@ -20,11 +21,14 @@ def get_date_value(self, object, field_name, request_date, period):
...
+
+        Returns:
+            A date object representation of the date after calculation.
         """
         if not getattr(object, field_name):
-            return request_date + relativedelta(years=period)
+            return datetime.strptime(request_date, "%Y-%m-%d").date() + relativedelta(years=period)
```
helrond, line 32: OK, this is fine for now, but let's remember that we're asking for a specific type of date formatting, which we need to check for somewhere.

helrond: Also, how about using `date.fromisoformat()` ? https://docs.python.org/3/library/datetime.html#datetime.date.fromisoformat

helrond (changes requested):

assign_rights/assemble.py
```diff
@@ -6,20 +9,47 @@
...
+    def calculate_dates(self, object, request_start_date, request_end_date):
+        """Calculate rights start and end dates for a given object.
+
+        Args:
+            object (obj): a RightsShell or RightsGranted object.
+            request_start_date (string): the start date for a group of records.
+            request_end_date (string): the end date for a group of records.
+
+        Returns:
+            object_start, object_end (list): a list with two datetime objects
```
helrond:
```suggestion
            object_start, object_end (tuple): a tuple with two datetime objects
```

helrond (changes requested):

assign_rights/assemble.py
p-galligan: @helrond @bonniegee As referenced in the latest commit, I'm using `date.fromisoformat()`, but that requires Python 3.7, so I've also updated the Dockerfile.

helrond: @p-galligan Ooh, I missed that detail! I believe we are running Python 3.6 on our servers, so we may need to go back to the previous implementation: 

```
datetime.strptime(request_date, "%Y-%m-%d").date()
```

Sorry about the confusion! That will teach me to read the documentation carefully...

assign_rights/assemble.py
```diff
@@ -6,20 +10,47 @@
...
+        return [RightsShell.objects.get(pk=ident) for ident in rights_ids]
+
+    def get_date_value(self, object, field_name, request_date, period):
+        """Calculates the value for a date.
+
+        Args:
+            object: the object for which a date is to be calculated.
+            field_name: the object attribute containing date data.
+            request_date: representation of a date sumbitted in a request.
+            period: the numbe of years to be used in calculating the date.
```
helrond:
```suggestion
            object (obj): the RightsShell or RightsGranted object for which a date is to be calculated.
            field_name (str): the object attribute containing date data.
            request_date (str): string representation of a date in ISO format.
            period (int): the number of years to be used in calculating the date.
```

helrond (changes requested):

helrond (approved):

HaSistrunk (approved):

PR #30, opened 2020-07-14
UIs for creating, editing, viewing, and browsing rights statement shells

bonniegee:
Adds UIs for creating, editing, viewing, and browsing rights statement shells (and associated acts). All fields are available to view and edit; additional validation dependent on rights basis will need to be added. A table was used in the rights shell detail view to show all fields, but that may not be the best way of presenting information.

### DIFF
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -25,7 +25,7 @@ from django.urls import path
 urlpatterns = [
     path('admin/', admin.site.urls),
     path('api/rights-assemble/', RightsAssemblerView.as_view(), name='rights-assemble'),
-    path('groupings/', GroupingListView.as_view(), name="groupings-list"),
+    path('', GroupingListView.as_view(), name="groupings-list"),
     path('groupings/<int:pk>/', GroupingDetailView.as_view(), name="groupings-detail"),
     path('groupings/create/', GroupingCreateView.as_view(), name="groupings-create"),
     path('groupings/<int:pk>/update/', GroupingUpdateView.as_view(), name="groupings-update"),
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -34,5 +34,5 @@ urlpatterns = [
     path('rights/', RightsShellListView.as_view(), name='rights-list'),
     path('rights/create/', RightsShellCreateView.as_view(), name='rights-create'),
     path('rights/<int:pk>/update/', RightsShellUpdateView.as_view(), name='rights-update'),
-    path('rights/<int:pk>', RightsShellDetailView.as_view(), name='rights-detail'),
+    path('rights/<int:pk>/', RightsShellDetailView.as_view(), name='rights-detail'),
 ]
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -28,7 +28,7 @@ urlpatterns = [
     path('groupings/', GroupingListView.as_view(), name="groupings-list"),
     path('groupings/<int:pk>/', GroupingDetailView.as_view(), name="groupings-detail"),
     path('groupings/create/', GroupingCreateView.as_view(), name="groupings-create"),
-    path('groupings/<int:pk>/edit/', GroupingUpdateView.as_view(), name="groupings-update"),
+    path('groupings/<int:pk>/update/', GroupingUpdateView.as_view(), name="groupings-update"),
     path('login/', AquilaLoginView.as_view(template_name="users/login.html"), name="login"),
     path('logout/', LogoutView.as_view(next_page="/login"), name="logout"),
     path('rights/', RightsShellListView.as_view(), name='rights-list'),
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -15,7 +15,9 @@ Including another URLconf
 """
 from assign_rights.views import (AquilaLoginView, GroupingCreateView,
                                  GroupingDetailView, GroupingListView,
-                                 GroupingUpdateView, RightsAssemblerView)
+                                 GroupingUpdateView, RightsAssemblerView,
+                                 RightsShellCreateView, RightsShellDetailView,
+                                 RightsShellListView, RightsShellUpdateView)
 from django.contrib import admin
 from django.contrib.auth.views import LogoutView
 from django.urls import path
@@ -29,4 +31,8 @@ urlpatterns = [
     path('groupings/<int:pk>/edit/', GroupingUpdateView.as_view(), name="groupings-update"),
     path('login/', AquilaLoginView.as_view(template_name="users/login.html"), name="login"),
     path('logout/', LogoutView.as_view(next_page="/login"), name="logout"),
+    path('rights/', RightsShellListView.as_view(), name='rights-list'),
+    path('rights/create/', RightsShellCreateView.as_view(), name='rights-create'),
+    path('rights/<int:pk>/update/', RightsShellUpdateView.as_view(), name='rights-update'),
+    path('rights/<int:pk>', RightsShellDetailView.as_view(), name='rights-detail'),
 ]
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -19,8 +19,8 @@ class RightsShellForm(ModelForm):
             'copyright_status',
             'determination_date',
             'note',
-            'applicable_start_date',
-            'applicable_end_date',
+            'start_date',
+            'end_date',
             'start_date_period',
             'end_date_period',
             'end_date_open',
@@ -35,6 +35,7 @@ class RightsGrantedForm(ModelForm):
         model = RightsGranted
         fields = [
             'basis',
+            'restriction',
             'act',
             'note',
             'start_date',
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -4,14 +4,12 @@ from .models import Grouping, RightsGranted, RightsShell
 
 
 class GroupingForm(ModelForm):
-    """docstring for RightsShellForm"""
     class Meta:
         model = Grouping
         fields = ["title", "description", "rights_shells"]
 
 
 class RightsShellForm(ModelForm):
-    """docstring for RightsShellForm"""
     class Meta:
         model = RightsShell
         fields = [
@@ -30,7 +28,6 @@ class RightsShellForm(ModelForm):
 
 
 class RightsGrantedForm(ModelForm):
-    """docstring for RightsShellForm"""
     class Meta:
         model = RightsGranted
         fields = [
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -1,10 +1,53 @@
-from django import forms
+from django.forms import ModelForm, inlineformset_factory
 
-from .models import Grouping
+from .models import Grouping, RightsGranted, RightsShell
 
 
-class GroupingForm(forms.ModelForm):
+class GroupingForm(ModelForm):
     """docstring for RightsShellForm"""
     class Meta:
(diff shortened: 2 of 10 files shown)


aquila/urls.py
```diff
@@ -29,4 +31,8 @@
     path('groupings/<int:pk>/edit/', GroupingUpdateView.as_view(), name="groupings-update"),
```
helrond:
```suggestion
    path('groupings/<int:pk>/update/', GroupingUpdateView.as_view(), name="groupings-update"),
```
Aligning naming conventions

aquila/urls.py
```diff
@@ -29,4 +31,8 @@
...
     path('login/', AquilaLoginView.as_view(template_name="users/login.html"), name="login"),
     path('logout/', LogoutView.as_view(next_page="/login"), name="logout"),
+    path('rights/', RightsShellListView.as_view(), name='rights-list'),
+    path('rights/create/', RightsShellCreateView.as_view(), name='rights-create'),
+    path('rights/<int:pk>/update/', RightsShellUpdateView.as_view(), name='rights-update'),
+    path('rights/<int:pk>', RightsShellDetailView.as_view(), name='rights-detail'),
```
helrond:
```suggestion
    path('rights/<int:pk>/', RightsShellDetailView.as_view(), name='rights-detail'),
```
I think you need a trailing slash here.

assign_rights/forms.py
```diff
@@ -1,10 +1,54 @@
-from django import forms
+from django.forms import ModelForm, inlineformset_factory
 
-from .models import Grouping
+from .models import Grouping, RightsGranted, RightsShell
 
 
-class GroupingForm(forms.ModelForm):
+class GroupingForm(ModelForm):
     """docstring for RightsShellForm"""
```
helrond: Update docstring

assign_rights/forms.py
```diff
@@ -1,10 +1,54 @@
...
     class Meta:
         model = Grouping
         fields = ["title", "description", "rights_shells"]
+
+
+class RightsShellForm(ModelForm):
+    """docstring for RightsShellForm"""
```
helrond: Update docstring

assign_rights/forms.py
```diff
@@ -1,10 +1,54 @@
...
+            'start_date_period',
+            'end_date_period',
+            'end_date_open',
+            'license_terms',
+            'statute_citation'
+        ]
+
+
+class RightsGrantedForm(ModelForm):
+    """docstring for RightsShellForm"""
```
helrond: Update docstring

assign_rights/templates/rights/manage.html
```diff
@@ -0,0 +1,14 @@
+{% extends 'base.html' %}
+
+{% block content %}
+
+<form action="." method="POST" accept-charset="utf-8">
+	{% csrf_token %}
+	{{ form.as_p }}
+      {{ rights_granted.as_p }} 
+	  {{ rights_granted.management_form }}
```
helrond:
```suggestion
      {{ rights_granted_form.as_p }}
      {{ rights_granted_form.management_form }} 
```
Would suggest a more descriptive variable name.

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
         return context
 
 
-def RightsShellListView(ListView):
+class RightsShellListView(LoginRequiredMixin, ListView):
```
helrond: Use `PageTitleMixin` in these views so we get a `h1` on these pages.

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
     '''browse and search rights shells'''
-    pass
+    queryset = RightsShell.objects.all()
```
helrond: You can also do 
```suggestion
    model = RightsShell
```
(see GroupingListView below) - whatever we choose we should just be consistent.

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+class RightsShellCreateView(LoginRequiredMixin, CreateView):
+    model = RightsShell
+    template_name = "rights/manage.html"
+    form_class = RightsShellForm
+    success_url = None
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if self.request.POST:
+            context['rights_granted'] = RightsGrantedFormSet(self.request.POST)
```
helrond:
```suggestion
            context['rights_granted_form'] = RightsGrantedFormSet(self.request.POST)
```

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+        else:
+            context['rights_granted'] = RightsGrantedFormSet()
```
helrond:
```suggestion
            context['rights_granted_form'] = RightsGrantedFormSet()
```

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+        return context
+
+    def form_valid(self, form):
+        context = self.get_context_data(form=form)
+        rights_granted = context['rights_granted']
```
helrond:
```suggestion
        rights_granted_form = context['rights_granted_form']
```

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+        if rights_granted.is_valid():
```
helrond, line 55:
```suggestion
        if rights_granted_form.is_valid():
```

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+            response = super().form_valid(form)
+            rights_granted.instance = self.object
```
helrond, line 57:
```suggestion
            rights_granted_form.instance = self.object
```

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+            rights_granted.save()
```
helrond, line 58:
```suggestion
            rights_granted_form.save()
```

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+
+    def get_success_url(self):
+        return reverse_lazy("rights-detail", kwargs={"pk": self.object.pk})
 
 
-class RightsShellDetailView(DetailView):
+class RightsShellDetailView(LoginRequiredMixin, DetailView):
     '''view a rights shell'''
-    pass
+    queryset = RightsShell.objects.all()
```
helrond: again, you can just do
```suggestion
    model = RightsShell
```

which in this case feels a little more intuitively right?

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+    template_name = "rights/manage.html"
+    form_class = RightsShellForm
+    success_url = None
+
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        if self.request.POST:
+            context['rights_granted'] = RightsGrantedFormSet(self.request.POST, instance=self.object)
+        else:
+            context['rights_granted'] = RightsGrantedFormSet(instance=self.object)
```
helrond:
```suggestion
            context['rights_granted_form'] = RightsGrantedFormSet(self.request.POST, instance=self.object)
        else:
            context['rights_granted_form'] = RightsGrantedFormSet(instance=self.object)
```

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+        return context
+
+    def form_valid(self, form):
+        context = self.get_context_data(form=form)
+        rights_granted = context['rights_granted']
+        if rights_granted.is_valid():
+            response = super().form_valid(form)
+            rights_granted.instance = self.object
+            rights_granted.save()
```
helrond:
```suggestion
        rights_granted = context['rights_granted']
        if rights_granted_form.is_valid():
            response = super().form_valid(form)
            rights_granted_form.instance = self.object
            rights_granted.save()
```

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
 
 
-class RightsShellUpdateView(UpdateView):
-    '''update rights shell'''
-    pass
+class RightsShellUpdateView(LoginRequiredMixin, UpdateView):
+    model = RightsShell
+    template_name = "rights/manage.html"
+    form_class = RightsShellForm
+    success_url = None
```
helrond:
```suggestion
```

This is unnecessary

assign_rights/views.py
```diff
@@ -28,24 +28,74 @@ def get_context_data(self, **kwargs):
...
+        if rights_granted.is_valid():
+            response = super().form_valid(form)
+            rights_granted.instance = self.object
+            rights_granted.save()
+            return response
+        else:
+            return super().form_invalid(form)
+
+    def get_success_url(self):
+        return reverse_lazy("rights-detail", kwargs={"pk": self.object.pk})
```
helrond:
```suggestion
```
This is also unnecessary.

assign_rights/views.py
helrond: The docstrings for these classes could also use a review (and should be enclosed in double quotes, not single quotes).

p-galligan (approved):

helrond (changes requested):
Overall this looks really good, and I appreciate the attention to using Django idioms, which allows you to write a lot less code!

I've added smaller comments inline but a few big rocks:
- We need tests! It may be possible to slightly tweak the tests I've added for Groupings to accommodate RightsShells (by adding an another for loop).
- I'm not a huge fan of the table in the detail view. I'm not sure how accessible it is, and it's honestly pretty hard to tell what you're looking at. I could see a table being used for the acts, but I'd lay that out vertically with proper table headers and all that.

bonniegee:
@helrond I don't love the table either, but I wasn't sure of a better way. It looks like Aurora uses the timeline functionality of Bootstrap, which I wasn't sure how to implement it. Would it be better to just go with fields surrounded by `<p>` tags for now?

helrond:
@bonniegee I think that timeline is an AdminLTE thing; it's not in Bootstrap. 

Either just plain `<p>` tags or a [definition list](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl) would work for now, I think.

HaSistrunk:
Nice! I'm seeing the need for page titles, which I think are implemented in the views using the `PageTitleMixin` and `page_title` attribute.

assign_rights/tests.py
```diff
@@ -50,6 +52,31 @@ def test_grouping_form(self):
                 form.errors[field], False,
                 "Field-specific error message not raised for {}".format(field))
 
+    def test_rightshell_views(self):
```
helrond, line 55: We can leave this for now,  but I think eventually we'll want to refactor these tests because they're basically the same for RightsShells and Groupings.

bonniegee: Cool--I also want to address rights granted or restricted in these tests in future iterations.

HaSistrunk (approved):

helrond (approved):
I just cleaned up some issues in the templates - looks like your editor may be allowing a mix of spaces and tabs, which makes things look messy in templates but could cause bigger problems in code blocks.

PR #33, opened 2020-07-14
Create JSON out of serialized data and calculated dates.

### LINKED ISSUE #4: Assemble rights statement
Combine calculated dates with other field content from rights shells in JSON that is returned to the service that sent a POST request.

p-galligan:
Fixes #4 . Assembles rights statements out of serialized JSON.


p-galligan:
Trying to figure out how to serialize objects to JSON, which is failing right now. @bonniegee @helrond do either of you have any experience with that?

As it is right now, I get this error from line 130 in the tests: `rest_framework.exceptions.ValidationError: {'non_field_errors': [ErrorDetail(string='Invalid data. Expected a dictionary, but got RightsShell.', code='invalid')]}`

I know it has to do with line 58 in assemblers, but not positive how to fix it.

assign_rights/assemble.py
```diff
@@ -52,9 +55,15 @@ def calculate_dates(self, object, request_start_date, request_end_date):
...
         return object_start, object_end
 
-    def create_json(self):
+    def create_json(self, obj, object_start, object_end):
         """docstring for create_json"""
-    pass
+        if obj.__class__.__name__ == "RightsShell":
+            serializer = RightsShellSerializer(obj, context={"start_date": object_start, "end_date": object_end})
+        else:
+            serializer = RightsGrantedSerializer(obj, context={"start_date": object_start, "end_date": object_end})
```
helrond: I would consider passing the serializer class as an argument to this function. That will save you a couple of lines and make this function more flexible.

assign_rights/assemble.py
```diff
@@ -52,9 +55,15 @@ def calculate_dates(self, object, request_start_date, request_end_date):
...
+        bytes = JSONRenderer().render(serializer.data)
+        data = json.loads(bytes)
```
helrond: I'm not 100% sure but I think you may need to decode the bytes string before using `json.loads()`

assign_rights/assemble.py
```diff
@@ -52,9 +55,15 @@ def calculate_dates(self, object, request_start_date, request_end_date):
             object, "end_date", request_end_date, object.end_date_period)
         return object_start, object_end
 
-    def create_json(self):
+    def create_json(self, obj, object_start, object_end):
         """docstring for create_json"""
-    pass
+        if obj.__class__.__name__ == "RightsShell":
+            serializer = RightsShellSerializer(obj, context={"start_date": object_start, "end_date": object_end})
```
helrond: I'm curious about this approach of passing additional context data into the serializer. It seems to me like you could just set those attributes on `obj` directly and then pass those into the serializer? Would keep your serializer a lot cleaner.

assign_rights/serializers.py
```diff
@@ -0,0 +1,56 @@
+from rest_framework import serializers
+
+from .models import RightsGranted, RightsShell
+
+
+class RightsGrantedSerializer(serializers.ModelSerializer):
+    """docstring"""
```
helrond: Update docstring

assign_rights/serializers.py
```diff
@@ -0,0 +1,56 @@
...
+            "act",
+            "restriction",
+            "start_date",
+            "end_date",
+            "note"
+        )
+
+
+class RightsShellSerializer(serializers.ModelSerializer):
+    """Serializes changes to the RightsGranted model to fit within RAC rights schema"""
```
helrond:
```suggestion
    """Serializes instances of the RightsGranted model to fit within RAC rights schema"""
```

assign_rights/tests.py
```diff
@@ -119,6 +119,21 @@ def test_calculate_dates(self):
         for granted in shell.rightsgranted_set.all():
             self.check_object_dates(granted, request_start_date, request_end_date)
 
+    def test_create_json(self):
+        shell = random.choice(RightsShell.objects.all())
+        request_end_date = random_date().isoformat()
+        request_start_date = random_date().isoformat()
+        start_date, end_date = self.assembler.calculate_dates(shell, request_start_date, request_end_date)
```
helrond: I'm not sure that this line is strictly necessary - the serializer should handle whatever dates you give it.

assign_rights/tests.py
```diff
@@ -119,6 +119,21 @@ def test_calculate_dates(self):
...
+        serialized = self.assembler.create_json(shell, start_date, end_date)
+
+        granted_list = []
+        for granted in shell.rightsgranted_set.all():
+            start_date, end_date = self.assembler.calculate_dates(granted, request_start_date, request_end_date)
+            granted_list.append(self.assembler.create_json(granted, start_date, end_date))
+        for item in granted_list:
+            serialized["rights_granted"].append(item)
```
helrond: I think you could potentially wrap the logic in lines 123-127 in a `for` loop with works against a list of the classes you want to test (RightsShell and RightsGranted) and just dispense with this code. After all, you're really testing whether the serialization works as expected, not the whole run method.

assign_rights/tests.py
```diff
@@ -119,6 +119,21 @@ def test_calculate_dates(self):
...
+        self.assertTrue(isinstance(serialized, dict))
```
helrond: Would definitely validate here!

helrond (changes requested):

assign_rights/tests.py
p-galligan: Do you mean validate against the RAC json schema?

bonniegee (approved):

helrond (approved):

HaSistrunk (approved):

PR #51, opened 2020-09-02
Create user in entrypoint.sh

### LINKED ISSUE #48: Create user in entrypoint.sh
Users are not automatically created after building the docker image; currently, you need to run the following in the Django shell:
```
docker-compose exec aquila-web python manage.py shell
User.objects.create_user("test", "test@example.com", "testpassword")
```

Create a default user in entrypoint.sh.

bonniegee:
Point to new script which creates a default user if there are no users when building the Docker image. It's possible this may not be desired behavior for deployment in our dev environment with SSO, but I wasn't sure of how to account for that.

This also changes the LOGIN_URL configuration in settings.py to point to a config file so that SSO can be bypassed while doing local development.

Fixes #48


helrond:
I don't see `setup_user.py` in the repo.

Otherwise, I think this approach makes sense. That function will only be executed when running `docker-compose up` so it won't interfere with other deployments, because those are not container-based.

setup_user.py
```diff
@@ -0,0 +1,6 @@
+from assign_rights.models import User
+
+if len(User.objects.all()) == 0:
```
helrond: this is a fine way to do it, but you could also check to see if that user already exists: 

`if User.objects.filter(username="test").exists():`

helrond (approved):
Just take care of linting and you should be good to go!

PR #57, opened 2020-09-21
Fix bug which only created test user if it already existed

### LINKED ISSUE #56: Building the application does not create test users
## Describe the bug
Building Aquila should add a test user with username: `test`, email: `test@example.com` and password: `testpassword`. Unfortunately this doesn't happen.

I see there is some functionality set up in `setup_user.py` but i don't think it's actually calling this and making a user.

## Expected behavior
Building the application should add the test users.

## Error message or screenshots
Here's the message when manually calling that file:

$ docker-compose exec aquila-web python setup_user.py
Traceback (most recent call last):
  File "setup_user.py", line 1, in <module>
  File "/code/assign_rights/models.py", line 3, in <module>
    from django.contrib.auth.models import AbstractUser
  File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/models.py", line 2, in <module>
    from django.contrib.auth.base_user import AbstractBaseUser, BaseUserManager
  File "/usr/local/lib/python3.6/site-packages/django/contrib/auth/base_user.py", line 47, in <module>
    class AbstractBaseUser(models.Model):
  File "/usr/local/lib/python3.6/site-packages/django/db/models/base.py", line 103, in __new__
    app_config = apps.get_containing_app_config(module)
  File "/usr/local/lib/python3.6/site-packages/django/apps/registry.py", line 252, in get_containing_app_config
    self.check_apps_ready()
  File "/usr/local/lib/python3.6/site-packages/django/apps/registry.py", line 134, in check_apps_ready
    settings.INSTALLED_APPS
  File "/usr/local/lib/python3.6/site-packages/django/conf/__init__.py", line 79, in __getattr__
    self._setup(name)
  File "/usr/local/lib/python3.6/site-packages/django/conf/__init__.py", line 64, in _setup
    % (desc, ENVIRONMENT_VARIABLE))
django.core.exceptions.ImproperlyConfigured: Requested setting INSTALLED_APPS, but settings are not configured. You must either define the environment variable DJANGO_SETTINGS_MODULE or call settings.configure() before accessing settings.

bonniegee:
Fixes #56 

Reverses logic in setup_user.py so that it checks if the test user does not exist before creating the account.

### DIFF
--- a/setup_user.py
+++ b/setup_user.py
@@ -1,5 +1,5 @@
 from assign_rights.models import User
 
-if User.objects.filter(username="test").exists():
+if not User.objects.filter(username="test").exists():
     print("Creating user")
     User.objects.create_user("test", "test@example.com", "testpassword")


helrond (approved):
Right!

p-galligan (approved):
Looks good!

HaSistrunk (approved):

Issue #35, opened 2020-07-14
Improve UI color contrast in styles


HaSistrunk:
Default Bootstrap styles don't always meet minimum accessibility standards. Check color contrast in UI and add custom styles to improve contrast where needed.

PR #58, opened 2020-09-22
Add logout link

### LINKED ISSUE #54: Add ability to log out from the application
Currently there is no way to log out of the application once logged in. 

Add the ability to logout. For now, this could be a link in the header.

p-galligan:
Add link to log out of application. Automatically redirects to login page. Fixes #54 

I think there's an accessibility/styling conversation to be had here. The contrast between the links in the navbar and the color of the header isn't great. We may want to introduce some custom styling.

### DIFF
--- a/assign_rights/templates/navbar.html
+++ b/assign_rights/templates/navbar.html
@@ -21,6 +21,11 @@
       <div class="form-inline my-2 my-lg-0">
         {% if request.user.is_authenticated %}
           <span class="navbar-text">Logged in as {{request.user.username}}</span>
+          <ul class="navbar-nav mr-auto">
+            <li class="nav-item">
+              <a class="nav-link" href="{% url 'logout' %}">Log Out</a>
+            </li>
+          </ul>
         {% endif %}
       </div>
     </div>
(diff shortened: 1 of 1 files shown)


HaSistrunk:
This is a little funky on mobile. I'm wondering if we might simplify into just one link that would indicate both that someone is logged in, and allow them to logout? Like:
```
      {% if request.user.is_authenticated %}
        <div class="navbar-nav my-lg-0">
          <span class="nav-item">
            <a class="nav-link" href="{% url 'logout' %}">Log out {{request.user.username}}</a>
          </span>
        </div>
      {% endif %}
```

Good call about the colors in the nav. I'm not a fan of that dark gray from Bootstrap either. I can look at some alternatives as part of issue #35 .

p-galligan:
Good idea. I've integrated the changes.

assign_rights/templates/navbar.html
```diff
@@ -20,12 +20,11 @@
       </ul>
       <div class="form-inline my-2 my-lg-0">
```
HaSistrunk, line 21:
```suggestion
```

HaSistrunk (approved):

helrond (approved):
Looks good to me!

In terms of contrast, the Bootstrap [color utilities](https://getbootstrap.com/docs/4.5/utilities/colors/) may be useful.

bonniegee (approved):
Looks good!

PR #60, opened 2020-09-22
Improve UI color contrast

HaSistrunk:
Fixes issue #35 

- Adjusts the button color and hover to pass WCAG AA contrast specs
- Lightens the nav link text for improved contrast

I also added the favicon image for a bonus reward.


assign_rights/templates/head.html
```diff
@@ -6,8 +6,10 @@
 <meta content="width=device-width, initial-scale=1, user-scalable=yes" name="viewport">
 <!-- Bootstrap 4.5.0 -->
 <link rel="stylesheet" href="{% static "bootstrap/css/bootstrap.min.css" %}">
+<!-- Custom styles-->
+<link rel="stylesheet" href="{% static "bootstrap/css/custom.css" %}">
```
helrond: I think I would prefer if this custom stylesheet were not nested within this bootstrap directory. That would make it easier to just replace the Bootstrap files if we want to upgrade.

p-galligan (approved):
Looks slick to me!

helrond (changes requested):

helrond (approved):

bonniegee (approved):
Looks great!

Issue #52, opened 2020-09-14
Validate rights form depending on basis


bonniegee:
## Is your feature request related to a problem? Please describe.
Different rights bases have different fields (e.g., copyright rights statements include jurisdiction, but this field does not apply to policy rights statements). To streamline the number of fields in the database, basis-specific validation does not happen in the rights model. Therefore, this validation needs to happen in the form.

## Describe the solution you'd like
Using the rights basis selected, make sure certain fields and values are present before saving.

## Additional context
See our local PREMIS guidelines for which fields and values are required: https://docs.rockarch.org/premis-rights-guidelines/guidelines/

PR #61, opened 2020-09-29
Show specific fields on rights form depending on basis

### LINKED ISSUE #53: Show specific fields on rights form depending on basis
Different rights bases have different fields (e.g., copyright rights statements include jurisdiction, but this field does not apply to policy rights statements). To streamline the number of fields in the database, different bases share fields.

Using the rights basis selected, only show fields applicable to that basis.

## Additional context
See our local PREMIS guidelines for which fields apply to which bases: https://docs.rockarch.org/premis-rights-guidelines/guidelines/

bonniegee:
This is very much a work in progress, but I'm getting stuck with both django-crispy-forms and Javascript/Vue.js.

I've been using these resources:

- Dynamic forms with Django and Vue.js https://medium.com/swlh/building-dynamic-forms-with-django-formsets-and-vue-js-f3c6e2dddd4a#81c0
- Vue.js in a Django template https://vsupalov.com/vue-js-in-django-template/
- Django crispy forms https://django-crispy-forms.readthedocs.io/en/latest/form_helper.html

I've done the following so far:

- In forms.py, I added a form for the basis field only and a form for copyright specific fields (if this is the right approach, the plan is to add forms for each of the bases). I'm not sure if these should be forms or formsets, or if there's another way of handling these. My idea was that the basis would always be rendered, and the other forms would be conditionally rendered.
- In views.py, I added references to both forms that I added.
- In base.html I added a line to include vue.js 
- In manage.html, I rendered the new forms and added a placeholder area for vue.js code

I think what I want to do is:
- Add a vue data binding to the rights basis field
- Depending on the user input to the rights basis field, render one of the basis-specific forms

I'm struggling with (at least) the following two things:
- Adding an attribute to a field in a crispy form
- Writing the Vue.js script to refer to that variable and then do something with it

### DIFF
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -27,6 +27,29 @@ class RightsShellForm(ModelForm):
         ]
 
 
+class BasisForm(ModelForm):
+    class Meta:
+        model = RightsShell
+        fields = [
+            'rights_basis',
+        ]
+
+
+class CopyrightForm(ModelForm):
+    class Meta:
+        model = RightsShell
+        fields = [
+            'copyright_status',
+            'determination_date',
+            'note',
+            'start_date',
+            'end_date',
+            'start_date_period',
+            'end_date_period',
+            'end_date_open',
+        ]
+
+
 class RightsGrantedForm(ModelForm):
     class Meta:
         model = RightsGranted
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -1,4 +1,4 @@
-from django.forms import ModelForm, inlineformset_factory
+from django.forms import ModelForm, Select, inlineformset_factory
 
 from .models import Grouping, RightsGranted, RightsShell
 
@@ -25,6 +25,7 @@ class RightsShellForm(ModelForm):
             'license_terms',
             'statute_citation'
         ]
+        widgets = {'rights_basis': Select(attrs={'v-model': 'selected'})}
 
 
 class RightsGrantedForm(ModelForm):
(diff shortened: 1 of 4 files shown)


bonniegee:
This is now working and fixes #53. However, I'm not sure if it could be DRYer.

I've done the following:
- Added the [v-model](https://v3.vuejs.org/guide/introduction.html#handling-user-input) attribute for the rights basis field in `forms.py`
- Included a Vue script in `rights/manage.py` to set a variable on the rights basis dropdown
- Used Vue if statements to show different fields based on user input in rights basis dropdown
- Use the `as_crispy_field` parameter in rendering fields to use the power of django-crispy-forms

There are placeholders for widgets for the additional fields that I ended up not needing for #53, but I've left them as placeholders because I branched off of this branch to work on #52.

helrond:
This all makes sense to me from looking at it. If you wanted to be DRYer, I think the easiest thing would be to create multiple forms in Django, and then show/hide based on that. That way you could take advantage of subclassing and model inheritance and keep your templates clean.

bonniegee:
Cool--I'm going to go with that approach. I'm also hoping that it will make basis-specific validation (for #52) a bit easier.

HaSistrunk (approved):

HaSistrunk:
This is done, right? It all seems to be functioning well.

p-galligan (approved):
Pulled down and looks good to me.

One thing I'm not positive about is whether we could use docstrings on `forms.py`. I generally understand what's happening, but I'd say best practice would be to put a single line docstring on these to explain what they're doing.

Open to other ideas on that though.

helrond:
I think docstrings in the forms is a good idea, especially since this requires some knowledge of PREMIS.

helrond (approved):

bonniegee:
Thanks all! I'm going to merge without the docstrings addition, because there will be specific things to include there as part of tackling issue #52 (and I'll include them in that PR).

Issue #44, opened 2020-08-03
Add user groups or some other mechanism of managing access


helrond:
## Is your feature request related to a problem? Please describe.
Now that we've implemented SSO, any user can get into Aquila and do things with just their windows username and password. This is non-optimal.

## Describe the solution you'd like
Add user groups or some other method of managing user access more granularly. 

## Additional context
This may mean we want to add additional features to the app, such as the ability for sys admins to add and remove users from groups. We'll also need to consider carefully what groups we want and what permissions should be assigned to each.

bonniegee:
I think there needs to be at least 3 groups of users:

- Users who can search and view groupings and rights statements
- Users who can search, view, create, and update groupings and rights statements
- Users who can search, view, create, update, and delete groupings and rights statements, and manage users

To your point about anyone being able to log in, I'm wondering if there should be a user default who can't view anything. What are your thoughts?

helrond:
Well, I'm not sure there's anything super confidential in this app, so I wouldn't be opposed to the base permission being the first one you describe above (read-only access). I think it's more when we get into create/update/delete actions that we'll likely want a bit more control.

p-galligan:
@bonniegee Can a user be part of more than one group at once? (I don't think it'd be necessary given what's outlined above, but it would slightly impact work)

bonniegee:
Good question! Let's go with no: a user can be part of only one group.

PR #62, opened 2020-10-01
Add user groups and restrict permissions based on groups

p-galligan:
Didn't get a ton of time to dig into this, but this is a WIP/proof of concept of restricting user access based on user groups. Right now, view functionality is just limited by being logged in, but create/edit are restricted to the edit group.

Adds template tags to check user groups and remove buttons in the templates depending on group access.

Groups have been added through a migration.

There's a `delete` group, but there really isn't anything for it to work against or restrict right now.

Build fails because I didn't get a chance to update view tests.


assign_rights/mixins/authmixins.py
```diff
@@ -0,0 +1,30 @@
...
+
+class UserMixin(LoggedInMixinDefaults, UserPassesTestMixin):
+    """Overwrites the mixin defaults"""
+    authenticated_redirect_url = reverse_lazy("login")
+
+    def test_func(self):
+        """Checks whether a user is a superuser in the application"""
+        if self.request.user.is_superuser:
+            return True
+        return False
```
helrond: Doesn't this mean that only superusers are going to be allowed to do things in the application?

helrond (changes requested):
I haven't pulled this down and tested, but this all looks right (and really well done!) to me, with the exception of the one thing I've noted.

assign_rights/mixins/authmixins.py
p-galligan: In my local testing, that wasn't happening, however I could have set something else up incorrectly. But the argument for it being there isn't great either, so I just removed that section for now since it wasn't doing anything. Application works as expected without it.

bonniegee:
This looks good! I've tried to take a crack at updating the tests as a way to dive into what's going on, and while everything related to checking a user's group looks good and make sense, I'm getting a bit stuck at assigning or updating a user's group.  Did you mostly follow the [official Django docs](https://docs.djangoproject.com/en/3.1/ref/django-admin/) to set this up or found another resource useful?

p-galligan:
So, by default, no one's in a group because everyone that joins should have view access. 

As for assigning a user to a group programmatically it gets a little more difficult. 

```
from django.contrib.auth.models import Group
user = User.objects.get(username='username')
group = Group.objects.get(name='group_name')
group.user_set.add(user)
```

You could also do `user.groups.add(group)` if you wanted to tackle it the other way around.

I didn't find great official documentation about programmatically dealing with user groups, but there's definitely some stuff out there.

assign_rights/tests.py
```diff
@@ -26,6 +26,7 @@ def setUp(self):
         self.api_factory = APIRequestFactory()
         self.factory = RequestFactory()
         self.user = User.objects.create_user("test_user", "test@example.com", "testpass")
+        self.group = Group.objects.get(name='edit').user_set.add(User.objects.get(username='test_user'))
```
helrond: This is a little byzantine. How about:
```suggestion
        self.group = Group.objects.get(name='edit')
        self.user.groups.add(self.group)
```
Alternatively, if you don't need to set the group to a variable you can also just do 
```suggestion
        self.user.groups.add(Group.objects.get(name='edit'))
```

bonniegee:
Noting so I don't forget--before this gets merged to development, setup_user.py should be modified to add the "test" user to the "edit" group.

bonniegee (approved):

helrond (approved):

HaSistrunk (approved):

helrond:
fixes #44

PR #70, opened 2020-10-09
Add json data fixtures

### LINKED ISSUE #68: Add fixtures for JSON from Zorya
Currently there are no fixtures to replicate JSON requests from Zorya to test against.

Add JSON samples from Zorya to test against.

bonniegee:
Fixes #68


fixtures/1.json
```diff
@@ -0,0 +1,5 @@
+{
+	"identifiers": "['1', '3']",
```
helrond: Hmm, this doesn't look right to me. I would think that these would be passed as arrays, rather than string representations of arrays:
```suggestion
	"identifiers": ["1", "3"],
```
It might also be worth figuring out whether those IDs will be strings or numbers.

helrond (commented):
See my comment about the identifier array.

fixtures/1.json
bonniegee: The result of
```for obj in Bag.objects.all():
            print('"identifiers": {}, "start_date": {}, "end_date": {}'.format(obj.rights_id, obj.start_date, obj.end_date))
```
in the Zorya tests is:
```"identifiers": ['9', '16'], "start_date": 1911-06-12, "end_date": 1966-07-21
"identifiers": ['9', '13'], "start_date": 1949-04-10, "end_date": 1980-06-09
```
I'm not sure of the best way of figuring out what the `post` function from the `requests` module is creating?

helrond: OK, it looks to me like:
- The rights_id field in Zorya is a JSONField, which means that the data that gets passed in will be stored as JSON. That also means there's no validation on what's being passed into that array - the ids could be numbers or strings. It may be worth accounting for both of these cases.
- If there's no validation in Zorya on whether or not it gets passed a list of rights_ids (if there's only one, is it still in a list? What if it gets passed something like "1, 2, 3"? You'll need to account for that here.
- What's coming back from the Zorya tests would likely be serialized in JSON as what I described above. 
- If you want to see what Zorya is sending I'd print out `resp.data` or maybe `resp.body` in the `retrieve_rights()` method.

helrond (approved):
👍

PR #71, opened 2020-10-27
Add fixtures for rights assembly view tests

### LINKED ISSUE #46: Write tests for the run method
The `run` method of the `RightsAssembler` does not have any tests associated with it.

Write tests for the `run` method.

Related to #34

bonniegee:
Tests run method through testing rights assembly view. Adds fixtures to test valid and invalid requests. Since fixtures are reliant on stable pks in the test database, these tests need to be inherit from the TransactionTestCase ([see Django docs](https://docs.djangoproject.com/en/dev/topics/testing/advanced/#django.test.TransactionTestCase.reset_sequences)). Because these are the only tests that need this functionality, I broke them into their own class from the other view tests, but I'm not sure if that was the best approach.

Additionally, removes django-nose (which was causing issues with the test database) and updates coverage configs.

Fixes #46


assign_rights/tests.py
```diff
@@ -19,6 +22,47 @@
...
+            self.assertEqual(response.status_code, 500, response.data)
+            self.assertEqual(response.data["detail"], "Error retrieving rights shell: RightsShell matching query does not exist.")
+
+        # RightsAssembler throws an exception
+        with open(join(invalid_data_fixture_dir, "no_start_date.json"), 'r') as json_file:
+            invalid_data = json.load(json_file)
+            request = self.factory.post(reverse("rights-assemble"), invalid_data, content_type='application/json')
+            response = RightsAssemblerView.as_view()(request)
+            self.assertEqual(response.status_code, 500, "Request should have returned a 500 status code")
+            self.assertEqual(response.data["detail"], "Request data must contain 'identifiers', 'start_date' and 'end_date' keys.")
```
helrond: You could cut this code in half if you wrapped this in a for loop:

```
for fixture_name, message in [("invalid_rights_id.json", "Error retrieving rights shell: RightsShell matching query does not exist."), ("no_start_date.json", "Request data must contain 'identifiers', 'start_date' and 'end_date' keys.")]:
...
```

helrond (changes requested):
Looks ✨  - just made one suggestion for cutting down on duplication in the tests.

helrond (approved):

Issue #64, opened 2020-10-01
Prevent rights basis type from changing on the edit page


bonniegee:
## Is your feature request related to a problem? Please describe.
On both the create and edit page, fields are rendered in the form depending on what rights basis is selected. If a rights basis type changes in the edit page, there could be issues relating to the different fields that display for each type and validation.

## Describe the solution you'd like
Disable rights basis drop down if editing, and not creating, a rights statement shell.

helrond:
This is definitely the simplest implementation. It should be possible, however, to deal with these validation issues, particularly since a lot of it is happening in Javascript. But you can wait and see whether user testing indicates a more complex approach is necessary.

Issue #65, opened 2020-10-02
Only show relevant fields for rights basis on detail page


bonniegee:
## Is your feature request related to a problem? Please describe.
While the edit and create pages dynamically show different fields for different rights bases, the rights shell detail page still shows all rights basis fields. 

## Describe the solution you'd like
The detail page should only show fields relevant to that basis.

PR #72, opened 2020-10-29
Prevent rights basis type from changing, only show relevant fields for rights basis on detail page

bonniegee:
Splits create and update templates so the update page only renders fields for the rights basis of the rights shell and disables editing the rights basis. In order to disable the rights basis field on the edit page, I created a new form to disable that field in a widget.

The code to render specific forms in the update view is pretty verbose, but I was having issues trying to make it more concise.


bonniegee:
Adding that this fixes #64 and #65

helrond:
One thing you could do to make the views a little more concise is factor out a `get_formclass` method which takes a rights basis and returns the form class you want, something like:

```
def get_formclass(self, rights_basis):
  if rights_basis == "Copyright":
    form_cls = CopyrightForm.
...
  return form_cls
```

Then you could use it in your views like:

```
cls = self.get_formclass(context["object"].rights_basis)
context["basis_form"] = cls(self.request.POST, instance=self.object)
```

That would cut down on doing the same logic twice at least.

You could also create a static dictionary that references form classes by rights bases:

```
FORMS = {"copyright": Copyright}
```
And you'd use that like

```
cls = FORMS[context["object"].rights_basis]
...
```

I'm not sure one is objectively better than the other; it's a matter of preference and what you think is easier to read.

bonniegee:
Outstanding issue with this PR: while the correct form type shows on the edit page, the edit page cannot be saved and is throwing a "this field is required" error for rights basis.

helrond:
@bonniegee this is ready for review. Since you submitted the PR I can't explicitly add you as a reviewer...

bonniegee (commented):
This looks good to me! The `form_cls` function makes a ton of sense. (I can't approve this since it's "my" PR.)

HaSistrunk (approved):
Looks good to me!

PR #77, opened 2021-02-22
Add policy and donor bases

### LINKED ISSUE #66: No current way to specify policy or donor basis
Technically, the donor rights basis and policy rights basis are actually "other" rights bases, with an "other rights basis" value of "donor" or "policy." Currently, there is no way of indicating that there are policy or donor bases.

"Donor" and "policy" should be added as rights bases types; the "other" form should be displayed when either are selected by a user creating a rights shells; and these should be serialized as "other" rights bases when creating rights statements.

## Additional context
"Donor" and "policy" are the only types of "other" rights bases outlined in local RAC policy. We may want to further extend this, but that is not necessary for the MVP.

bonniegee:
Adds policy and donor basis to the rights shell model, and loads the "other" form when either is selected in create and edit pages. Fixes #66.


assign_rights/views.py
```diff
@@ -96,6 +96,10 @@ def get_form_cls(self, rights_basis):
             form_cls = StatuteForm
         elif rights_basis == "License":
             form_cls = LicenseForm
+        elif rights_basis == "Policy":
+            form_cls = OtherForm
+        elif rights_basis == "Donor":
+            form_cls = OtherForm
```
helrond: I think we can just let these additional cases get caught in the `else`, right?

helrond (changes requested):
When a user selects "Other" don't they then need to say what the "other" is? Or am I misremembering PREMIS?

One minor comment below.

bonniegee:
@helrond Other rights basis does require a specific value, but it needs to be from a controlled vocabulary in the repository. I'm inclined to not show "other" since we do not use any additional values of "other" locally, and maybe (in the future) cut it from the model. What do you think?

helrond:
I think I would lean towards cutting it out of the model now if we're not going to show it. It can always be added back in, but I'm not a huge fan of having code lying around that we're not using.

It might also be good to drop a docstring in the relevant model explaining this thinking so that future us doesn't forget when we need to add another type for `other`...

assign_rights/views.py
bonniegee: You're right 🤦‍♀️

helrond (approved):

HaSistrunk (approved):

PR #82, opened 2021-02-23
Add or remove rights granted

### LINKED ISSUE #63: Add multiple (and delete) rights granted or restricted
A rights basis can have 0 or more acts (rights granted or restricted) associated with it. The app currently does not support not having any acts, and it also does not allow for multiple acts to be added to a basis.

Add the ability to save a rights basis without an act, add multiple acts, and delete acts.

helrond:
Allows users to add multiple rights granted to a given rights basis, and to delete them as well. Also combines the `create` and `edit` templates into a single `manage` template to reduce code duplication.

fixes #63


HaSistrunk:
Cool! Two things:
1. The Rights Granted h3 (eg. Rights Granted 1) should increment for each new Rights Granted section, right? Right now they are all designated as 1.
2. Is there any way to make the Add New Rights Granted link a button? It has role=button, which is cool, but because it's a link the spacebar won't activate it which would be the expected pattern for a button.

bonniegee (commented):
This overall looks good and makes sense to me! The one thing I'm unsure about is that an additional blank act appears when editing a rights shell. For example, if I have a saved rights shell with 1 act, I see 2 total acts after clicking the edit button (the one I had saved and a new, blank one). I think I would err on starting with zero acts on the manage page if that's a necessary compromise, but maybe this is just something to deal with after user tests?

bonniegee:
I meant to include: should `rights/edit.html` be deleted?

helrond:
Good catches - made both of those changes.

HaSistrunk (approved):

bonniegee (approved):

Issue #80, opened 2021-02-23
Improve str method for RightsShells


helrond:
## Is your feature request related to a problem? Please describe.
The `str` method determines the string that is displayed for model instances in multi-select fields, such as the multi-select for RightsShells in the Groupings create and edit forms. For rights shells, this is currently set to `"{} ({})".format(self.note, self.rights_basis)`, which does not always provide the necessary information.

## Describe the solution you'd like
Provide a better display string. Consider using additional basis information as well as act/restriction information.

## Additional considerations
We may want to make additional fields required in the rights basis and/or rights granted model. In particular, it seems like the note could be a potentially useful field in helping humans disambiguate rights statements and we may want to make that required even though it's optional in PREMIS.

bonniegee:
I think in addition to, or as a substitute for, note content, the rights granted associated with a basis would be helpful in disambiguating rights statements. However, as there can be zero, one, or multiple acts associated with a basis, I'm not sure if that's compatible with the `__str__` method?

helrond:
That could work - the `__str__` method would just have to account for those cases.

PR #84, opened 2021-02-23
Clean up templates

### LINKED ISSUE #81: Improve display on RightsShell detail pages
The display of individual RightsShells is pretty hard to parse.

Implement a visual hierarchy and consider whether it is necessary for all fields to appear.

helrond:
Makes list views more concise by adding buttons to header. Adds `__str__` method to RightsGranted. Tightens up RightsShell detail view. Fixes #81


helrond:
Tests are failing here for the same reason they were failing in #82 - changes to the choices in `RIGHTS_BASIS_CHOICES` were not carried over to the tests. I've addressed that in #82, so once that is merged I'll pull in those changes here.

bonniegee (approved):
Looks great!

HaSistrunk:
Cool! Two little things I noticed:
1. Maybe a nit pick, but it looks crowded at the top to my eye. What do you think about increasing the margin on the `nav` to like mb-5 instead of mb-2?
2. We need a visual indication other than color for the title and ID values in the Groupings and Rights Shells tables to show that they are links. Maybe just add an underline? I think the ID links being such a small clickable area as numbers may be something to look out for in usability testing: do folks notice them, and do they understand where those links will take them?

helrond:
I've added additional padding to the navbar and added an underline to links. I agree that the ID is probably not the best link text (and not a big enough target). My proposal would be to wait for #80 and then use that as the link text. That may necessitate other changes in the table as well.

Additionally, we could look at a different HTML structure (divs rather than tables) so we could link the entire row. I'd suggest waiting for some user feedback first though.

HaSistrunk (approved):

PR #86, opened 2021-02-24
Improve str method for RightsShells, make note required for basis

bonniegee:
Builds out `__str__` method for rights basis, including adding applicable dates. Keeps `note` field  in method and modifies RightsShell model so that note field is required.

While the note field should always be included (to @helrond's point in the issue, and why I did ultimately make it required), I am concerned that the note content could be very long. This might be a thing to keep an eye out for in the future.


helrond:
OK I did some more thinking about this based on @bonniegee's comments and tweaked this PR a bit:
- Pulled `note` out of the `__str__` method, which meant that I could make it non-required again.
- Used some additional information from the rights basis form in cases where that information is likely to be available
- Use this str method in templates.

bonniegee:
I'm not sure if jurisdiction (for the relevant bases) adds much, since I think everything will be the US. That minor point aside, I think this looks great!

helrond:
OK cool, thanks @bonniegee. @HaSistrunk this is ready for your review!

HaSistrunk (approved):

PR #87, opened 2021-02-24
Debugs issues with formsets

### LINKED ISSUE #85: Drop downs are empty for acts
## Describe the bug
When adding a rights granted on the create page for a new rights shell, it is not possible to select an act or restriction from the drop down, as all there are no options.

## To reproduce
1. Create a new rights shell 
2. Add a new rights granted
3. Select the act or restriction drop-down

## Expected behavior
The drop downs contain the values set for each field from the rights granted model.

helrond:
I started out to address #85, but this turned into a bit of a journey. So, there are actually a couple of things this PR does.
1. allow deletion of formset instances (by flipping the boolean `can_delete`)
2. debug an error in the `addNewRightsGranted` function which was preventing formset indexes from being correctly calculated
3. Added `act` and `restriction` choices into the view so we can access those in the template. Previously I was pulling those from the first form in the formset, but because we've now set the `extra` parameter to `0` there are no forms in the formset. This solution feels a little kludgy, but it will work for now. fixes #85 
4. Debug some conditionals in the Rights Shell detail template.

### DIFF
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -107,5 +107,5 @@ RightsGrantedFormSet = inlineformset_factory(
     RightsGranted,
     extra=0,
     form=RightsGrantedForm,
-    can_delete=False
+    can_delete=True
 )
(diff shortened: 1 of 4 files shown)


bonniegee (approved):
This works great, and the changes in the code make sense to me.

HaSistrunk (approved):

PR #90, opened 2021-02-25
Allow users to delete Groupings and RightsShells

### LINKED ISSUE #83: Allow authorized users to delete rights shells and groupings
Users belonging to the `delete` group should be able to delete RightsShells and Groupings. There are currently no UI elements that support this.

Add delete views, authentication mixins and code in templates as necessary to support deletion of objects.

helrond:
Allows users in the `delete` group to delete Grouping and RightsShell instances.
- Adds delete views for Groupings and Rights Shells, and uses the `DeleteMixin` to ensure only users in the `delete` group have access to these views
- Add delete views to URL config
- Adds delete button to detail template and confirm templates

fixes #83

### DIFF
--- a/aquila/urls.py
+++ b/aquila/urls.py
@@ -17,8 +17,8 @@ from assign_rights.views import (AquilaLoginView, GroupingCreateView,
                                  GroupingDeleteView, GroupingDetailView,
                                  GroupingListView, GroupingUpdateView,
                                  RightsAssemblerView, RightsShellCreateView,
-                                 RightsShellDetailView, RightsShellListView,
-                                 RightsShellUpdateView)
+                                 RightsShellDeleteView, RightsShellDetailView,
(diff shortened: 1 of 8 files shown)


HaSistrunk:
Do I need to login as a user other than "test" to see this in the UI?

helrond:
The quickest way to do this is to remove the volume (`docker-compose down -v`) and then restart.

Longer explanation: the `test` user needs to be associated with the `delete` group in order for you to see these buttons/delete things. I've updated the `setup_user.py` script to do that, but that script only creates the user and adds it to groups if it doesn't exist already. So by removing the volume you're deleting the database and starting over, which means the user will be recreated with the desired groups. You could also get in the Django shell and add the user to the group manually.

HaSistrunk (approved):
This all looks good to me!

bonniegee (approved):
🙌

PR #92, opened 2021-02-25
Improves rights basis templates

### LINKED ISSUE #74: Improve display of required form fields
Required form fields display inconsistently. Some have asterisks while others do not.

Required fields should be consistently styled.

## Additional context
Initial investigation - the crispy forms filter does not seem to be picking up required fields from the Model Form. We may need to use the `crispy` tag: https://django-crispy-forms.readthedocs.io/en/latest/crispy_tag_forms.html

helrond:
Uses Crispy Layouts to improve rights basis form display. Overrides default field display to add required styles that have a required widget attribute (as opposed to only those fields which are marked as required on the model).

Fixes #74

### DIFF
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -36,6 +36,7 @@ class RightsShellForm(ModelForm):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.helper = FormHelper(self)
+        self.helper.field_template = "forms/custom_field.html"
         self.helper.form_tag = False
         self.helper.disable_csrf = True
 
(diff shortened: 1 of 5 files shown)


HaSistrunk:
I'm wondering if the start and end date fields should be date inputs instead of text inputs in the rights basis (like they are in the Rights Granted)?

bonniegee (approved):
I think this looks really great! I agree that there should be more discussion about date fields, but that can probably happen in #93

helrond:
@HaSistrunk agree - those date fields should be consistent. I was having trouble figuring out how to make that happen AND was also thinking that we may want to use the datepicker approach we've used other places (like DIMES and the library search).

HaSistrunk (approved):

Issue #67, opened 2020-10-06
Add validation for dates


bonniegee:
## Is your feature request related to a problem? Please describe.
In local practice, PREMIS rights bases and acts must include a start date, and optionally an end date. Since dates can either be a hard coded date (e.g., 1/1/1974) or a number of years after creation of the item they're being applied to (or, for end dates only, "open"), this requirement is not being enforced at the model level and should be enforced when the user is submitting a form.

## Describe the solution you'd like
Before a rights shell is saved, make sure that the following conditions are met for basis dates and any act dates:

- There is one and only one start date, either a calendar date or a number of years after creation
- There is no more than one end date, including the "open end date" boolean
- The end date is not before the start date

## Additional context
This likely needs to be done in Vue in manage.html and not the forms, but I'm not 100% certain of the best approach.

bonniegee:
I'm trying to do the validation in Vue, but getting stuck on 1) how to do the actual validation (in terms of when validation happens, and what happens with the errors) 2) how to deal with Vue variables (I did get as far as adding the `v-model` attributes to the relevant form field widgets). I have this method so far (modeled on [this example](https://vuejs.org/v2/cookbook/form-validation.html)), but I'm not sure of the cleanest way of dealing with complex validation in Vue/JS methods:
```
checkStartDates: function (e) {
  if ((!this.start_date && this.start_period) || (this.start_date && !this.start_period)) {
    return true;
  }

  if (!this.start_date && !this.start_period) {
    this.errors.push('Must have a start date or period.');
  }
  if (this.start_date  && this.start_period) {
    this.errors.push('Must have a start date or period.');
  }

}
```
@helrond Do you have any thoughts on the best approach?

helrond:
In terms of how to wire things up, this looks like a useful primer: https://vuejs.org/v2/cookbook/form-validation.html. It looks like there's just a handler on the form submit.

I think you're on the right track with the validation. My read is that you're getting stuck with "Only one of start_date and start_date_period"? This looks relevant: https://codereview.stackexchange.com/questions/199853/function-to-test-if-exactly-one-of-three-parameters-is-truthy-without-using-equ and I think you could do

```
if !(!!this.start_date + !!this.start_period == 1) {
  this.errors.push('Must have only one of either a start date or start period')
}

bonniegee:
[...]

PR #93, opened 2021-03-01
Improve user interaction pattern for date inputs

### LINKED ISSUE #91: Improve display of date fields
The current display of date fields is potentially really confusing to a user. Are there ways we can improve the interaction pattern or display to make things easier on users? 

Screen Shot 2021-02-25 at 12 01 29 PM

Potential solution:
- Instead of displaying all the date inputs, display a select for both start date and end date. 
- For start date, the options would be:
  - "This restriction starts on a specific date" - if selected, displays the start_date field
  - "This restriction starts after an embargo period" - if selected, displays the start_date_period
- For end date, the options would be:
  - "This restriction ends on a specific date" - if selected, displays the end_date field
  - "This restriction ends after an embargo period" - if selected, displays the end_date_period
  - "There is no end date for this restriction" - doesn't display any input but the end_date_open boolean would be checked

One advantage to an approach like this is that it would make validation a lot easier. 

Are there accessibility implications of this kind of pattern?

## Additional context
Noting this was also an issue identified (but not fully solved) in Aurora testing.

helrond:
Adds date inputs as specified in #91. This should be a little easier for users to work with, and substantially simplifies the UI as well as validation.

I'm interested in whether having some form layouts in Crispy Form Layouts and some in regular Django templates is confusing/causes maintenance issues. If so, I think I would argue for us just dropping everything into Django templates, and shedding crispy as a dependency.

Not covered in this PR - date inputs are not consistently structured, and behave differently. We should fix that, see #99

This also presumes that #98 will be merged first.

fixes #91


bonniegee:
I think this approach looks really great, and it also obviates (or in a way, fixes) #67.

One buggy behavior I've noticed is that when a Start Date Embargo Period is 0 (and the record is saved), that start date totally disappears when going to the edit page.

helrond:
@bonniegee thanks - let me look at that. 

In terms of validation, there are certain scenarios this won't cover, like an end date being after a start date. But I think it will take care of the basic "is there one and only one value here" validation.

bonniegee:
I'm trying to think through whether having the end date before the start date would cause issues (i.e., cause an app to blow up) downstream--I think the answer is "no," but it might be worth keeping this scenario in mind as we work on back end development.

helrond:
OK - I mean, let's do what we need to do so we get good data! All I'm saying is that I think #67 may need to be rescoped.

bonniegee (approved):
This looks ✨! 

Something to keep an eye out for in usability testing--I think the "embargo" language could possibly be confusing for copyright bases.

HaSistrunk (approved):

PR #98, opened 2021-03-03
Store rights bases as lowercase

### LINKED ISSUE #94: Should rights bases be lowercase?
The database and display values for rights bases are both title-cased. Unless there's a good reason for it, the value we store in the database should *probably* be normalized i.e. lowercase.

If this wasn't a conscious choice, change the first value of the `RIGHTS_BASIS_CHOICES` tuple to lowercase. This will also require a new migration.

If there's a reason for leaving as-is, let's leave it alone!

bonniegee:
Changes values of rights basis choices that are stored in the database to lowercase. Updates the Vue code in manage.html and relevant test to account for change.

Fixes #94


helrond (approved):

helrond (changes requested):
Oh wait you need to tweak the `get_form_cls` method in `RightsShellUpdateView` too...that's using the uppercase versions of those bases as well...

helrond:
Also possibly you could consider writing a [data migration](https://docs.djangoproject.com/en/3.1/topics/migrations/#data-migrations) to lowercase any existing rights shells. Definitely not a hard requirement but if you're interested in learning about how to do it would be a good exercise!

The `0005_auto_20200929_1724` migration in this app is an example of a data migration.

helrond (approved):

HaSistrunk (approved):

PR #102, opened 2021-03-09
Use text inputs with patterns

### LINKED ISSUE #99: Make date inputs look and behave consistently between rights basis and rights granted forms
For Reasons which have to do with crispy forms, etc. the date inputs on rights basis forms are different from those in rights granted forms. Additionally, these are plain `text` inputs in the rights basis forms, but are `date` inputs in rights granted forms.

Date inputs should be of the same type, and should behave the same way. I'd suggest just having plain `text` inputs and adding a custom datepicker.

Make these both `date` inputs and rely on browser datepickers.

## Additional context
This PR from the library discovery environment may be a useful reference: https://github.com/RockefellerArchiveCenter/library.rockarch.org/pull/88

helrond:
Converts inputs to text inputs and adds pattern validation.

I went around the block a couple of times on this one, and ended up back at a pretty simple solution. My reasoning is:
- Adding a datepicker introduces a fair amount of additional code and imports, between JQuery UI (CSS and JS) as well as the datepicker JS.
- Because the inputs it's attached to start out hidden, the datepicker needs to be called on certain DOM actions, which introduces even more complexity to the Vue app.
- This solution uses HTML attributes for validation, which should be handled properly from an accessibility perspective by all major browsers. It's not super beautiful, but it gets the job done.
- Given the potential dates in play, my guess is that a date picker is not going to be all that useful because users will have to navigate back many years. My guess is that most users will prefer to type the date in directly.

Fixes #99

### DIFF
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -15,12 +15,12 @@ class RightsShellCommonLayout(Layout):
                 Div("determination_date", css_class="col"), css_class="row"),
             Div(
                 Div(Field("rights_begin", v_model="rightsBegin"), css_class="col-5"),
-                Div("start_date", css_class="col-3", v_if="rightsBegin=='start_date'"),
-                Div("start_date_period", css_class="col-3", v_if="rightsBegin=='start_date_period'"), css_class="row"),
+                Div(Field("start_date", pattern=r"\d{4}-\d{2}-\d{2}"), css_class="col-4", v_if="rightsBegin=='start_date'"),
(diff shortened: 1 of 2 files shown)


p-galligan (approved):
Looks good! And I agree that a date picker would be more tedious than anything here, especially if you have to go back 60 years.

bonniegee (approved):
This looks great! I agree with the logic here, including that a date picker may be less useful for dates decades in the past. I think given the fact that a small number of users will be using this app, I think this is the right amount of constraints and validation.

PR #103, opened 2021-03-09
Focus on h3 of newly added rights granted

### LINKED ISSUE #101: Include accessible indication that a new rights granted section has been added to the form
When a user clicks the "Add New Rights Granted" button, the new rights granted section appears above the button while focus remains on the button. There should be some indication that something new has been added to the page for users who won't see the visual change.

After the user clicks the button, can we programmatically shift focus to the h3 of the newly added rights granted heading? Adding `tabindex=-1` to the h3 would enable focus (so that the heading won't be in the natural tab order, but the element can still be focused).

helrond:
Improves accessibility by shifting focus to the h3 of the newly added rights basis. 

I've used a tabindex of `-2` because there are bootstrap styles which remove the outline for `tabindex=-1`.

Fixes #101

### DIFF
--- a/assign_rights/templates/rights/manage.html
+++ b/assign_rights/templates/rights/manage.html
@@ -47,7 +47,9 @@
 
       <div class="card mb-4">
         <div class="card-header">
-          <h3 class="float-left">Rights Granted [[getIndexAmongVisibleRightsGranted(rights_granted_index) + 1]]</h3>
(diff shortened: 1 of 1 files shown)


p-galligan (approved):
I'm still kind of new to accessibility and focus, but the tab index usage seems correct to me.

bonniegee (approved):

HaSistrunk:
Great! Gave it a quick spin with VoiceOver to double check and this works nicely.

HaSistrunk (approved):

PR #104, opened 2021-03-09
Improves form validation

### LINKED ISSUE #96: Improve form error indications for rights basis
If there is an error in the form (like the start date is in the wrong format), and the user tries to save, they have to restart the form to see where the error occurred.

Given the approach we're using, I'm not sure what the best solution here is. At minimum, a message indicating that their were errors preventing submission that the user can see when they try and  fail to save would reduce confusion.

### LINKED ISSUE #100: Form validation should include making sure that start dates are before end dates
Currently, there is no check in form validation on whether the end date is before the start date. This is a possible user error that should be caught before the form is saved.

Before a rights shell is saved, make sure that the following condition is met for basis dates and any act dates:

- The end date is not before the start date

I think this should only compare the start and end if they're of the same type (i.e., a number of years or a date).

## Additional context
This is rescoped from #67.

helrond:
Improves form validation overall:
- add `required` attributes to required inputs
- improve date display for valid and invalid forms with templatetag
- add custom validation for start and end date, fixes #100 
- displays error messages in form, fixes #96 

Some additional housekeeping
- Add missing docstrings to tests
- Improve consistency of labels across views/objects

### DIFF
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -15,12 +15,12 @@ class RightsShellCommonLayout(Layout):
                 Div("determination_date", css_class="col"), css_class="row"),
             Div(
                 Div(Field("rights_begin", v_model="rightsBegin"), css_class="col-5"),
-                Div(Field("start_date", pattern=r"\d{4}-\d{2}-\d{2}"), css_class="col-4", v_if="rightsBegin=='start_date'"),
-                Div("start_date_period", css_class="col-4", v_if="rightsBegin=='start_date_period'"), css_class="row"),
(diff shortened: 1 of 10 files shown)


p-galligan (approved):

HaSistrunk:
Nice!
Should the start/end date and start/end date embargo labels include the "*" as well?

To confirm: Determination data is not required?

HaSistrunk (approved):

bonniegee:
When submitting a form that has the start date after the end date, a blank form is loaded and I don't see any errors. Is this the intended behavior?

helrond:
Nope, definitely not! Just pushed some changes that should fix that.

bonniegee (approved):
Seeing the errors now--looks great!

Issue #49, opened 2020-08-06
Send JSON back to an external app


bonniegee:
**Is your feature request related to a problem? Please describe.**
Aquila does not yet send JSON back to an application it receives a POST request from.

**Describe the solution you'd like**
Return rights JSON. There is a placeholder in the `return_rights` method.

Issue #47, opened 2020-08-06
Validate rights JSON in tests


bonniegee:
**Is your feature request related to a problem? Please describe.**
This application needs to reliably create valid rights JSON according to our local schema. 

**Describe the solution you'd like**
Include validation of rights JSON against our local schema in tests.

PR #109, opened 2021-03-10
Send JSON back to an external app

bonniegee:
This fixes the basic requirement of #49 by returning the `shell_data` array in the `run` method. Anything that would be handled the placeholder `retrieve_rights` method is already in the `run` method. There are some issues with the content of the `rights_data` that will require new issues (this is related to #47).


helrond (approved):

p-galligan (approved):
Looks good to me.

PR #112, opened 2021-03-15
Build out rights shell creation related methods in test helpers

### LINKED ISSUE #111: add_rights_shells method in test helpers produces invalid rights shells
The `add_rights_shells` method in the test helpers produces rights shells that are invalid (e.g., allows policy bases to have a jurisdiction, or allows an end date and an open end date). In order to have better tests, this data should hew more closely to what will be created in the app.

Build out the `add_rights_shells` method to create rights shells that adhere to PREMIS standards (and are the equivalent of what would be created through the UI).

bonniegee:
Makes three changes so that rights shells and acts created as part of setting up tests more closely mimic what would be created by users using the forms.

1. Builds out `random_date` method to take arguments, so that we can more closely target start dates and end dates.
2. Adds a `set_dates` method so that rights shells and acts only have one start date and one end date.
3. Create separate methods for different types of rights bases, so that only the correct fields are associated with each.

It looks like tests are passing on the latest commit, but when I ran tests locally these changes caused the `test_calculate_dates` method to fail. I documented this in #113; fixing it requires some changes in assembler.py as well as the tests, which will happen in a PR branched off of this.

Fixes #111

### DIFF
--- a/assign_rights/test_helpers.py
+++ b/assign_rights/test_helpers.py
@@ -67,9 +67,9 @@ def add_rights_shells(count=15):
             new_shell.jurisdiction = "us"
         if basis == "copyright":
             new_shell.copyright_status = random.choice(["copyrighted", "public domain", "unknown"])
-        if basis == "statute":
+        elif basis == "statute":
             new_shell.statute_citation = random_string()
-        if basis == "license":
+        elif basis == "license":
             new_shell.license_terms = random_string()
         new_shell.save()
 
--- a/assign_rights/test_helpers.py
+++ b/assign_rights/test_helpers.py
@@ -60,17 +60,18 @@ def set_dates():
 def add_rights_shells(count=15):
     for x in range(count):
         start_date, end_date, start_date_period, end_date_period, end_date_open = set_dates()
-        new_shell = RightsShell.objects.create(determination_date=random_date(10, 0), note=random_string(), start_date=start_date, end_date=end_date, start_date_period=start_date_period, end_date_period=end_date_period, end_date_open=end_date_open)
+        new_shell = RightsShell(determination_date=random_date(10, 0), note=random_string(), start_date=start_date, end_date=end_date, start_date_period=start_date_period, end_date_period=end_date_period, end_date_open=end_date_open)
         basis = random.choice(["copyright", "policy", "donor", "statute", "license"])
         setattr(new_shell, "rights_basis", basis)
         if basis in ["copyright", "statute"]:
-            setattr(new_shell, "jurisdiction", "us")
+            new_shell.jurisdiction = "us"
         if basis == "copyright":
-            setattr(new_shell, "copyright_status", random.choice(["copyrighted", "public domain", "unknown"]))
+            new_shell.copyright_status = random.choice(["copyrighted", "public domain", "unknown"])
         if basis == "statute":
-            setattr(new_shell, "statute_citation", random_string())
+            new_shell.statute_citation = random_string()
         if basis == "license":
-            setattr(new_shell, "license_terms", random_string())
+            new_shell.license_terms = random_string()
+        new_shell.save()
 
 
 def add_rights_acts(count=5):
--- a/assign_rights/test_helpers.py
+++ b/assign_rights/test_helpers.py
@@ -45,10 +45,10 @@ def set_dates():
     if choice == 1:
         start_date = random_date(100, 25)
         end_date = random_date(24, -50)
-    if choice == 2:
+    elif choice == 2:
         start_date = random_date(100, 25)
         end_date_period = random.randint(20, 75)
-    if choice == 3:
+    elif choice == 3:
         start_date_period = random.randint(0, 10)
         end_date_period = random.randint(20, 75)
     else:
--- a/assign_rights/test_helpers.py
+++ b/assign_rights/test_helpers.py
@@ -34,31 +34,103 @@ def add_groupings(count=5):
                 grouping.rights_shells.add(random.choice(RightsShell.objects.all()))
 
 
+def set_dates():
+    """Returns start and end dates in a way consistent with what is allowed in forms"""
+    start_date = None
+    end_date = None
+    start_date_period = None
+    end_date_period = None
+    end_date_open = False
+    choice = random.randrange(1, 5)
+    if choice == 1:
+        start_date = random_date(100, 25)
+        end_date = random_date(24, -50)
+    if choice == 2:
+        start_date = random_date(100, 25)
+        end_date_period = random.randint(20, 75)
+    if choice == 3:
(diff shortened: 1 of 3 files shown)


assign_rights/test_helpers.py
```diff
@@ -28,31 +34,103 @@ def add_groupings(count=5):
...
+    else:
+        start_date_period = random.randint(0, 10)
+        end_date_open = True
+    return start_date, end_date, start_date_period, end_date_period, end_date_open
+
+
 def add_rights_shells(count=15):
-    for x in range(count):
+    def copyright_shell():
+        date_info = set_dates()
```
helrond: Wondering if it makes sense to call this at line 121 and then pass the result as an argument to these various `*_shell` methods?

assign_rights/test_helpers.py
```diff
@@ -28,31 +34,103 @@ def add_groupings(count=5):
...
+        end_date_period = random.randint(20, 75)
+    else:
+        start_date_period = random.randint(0, 10)
+        end_date_open = True
+    return start_date, end_date, start_date_period, end_date_period, end_date_open
+
+
 def add_rights_shells(count=15):
-    for x in range(count):
+    def copyright_shell():
```
helrond: There's something about nested functions that doesn't feel quite right to me, and it also feels like there's a lot of duplication of code because of the shared fields between all these rights basis. 

How about creating a rights basis instance, and then adding additional fields by randomly choosing from a list which contains tuples or dicts of the fields for each rights basis?

helrond: There's something about nested functions that doesn't feel quite right to me, and it also feels like there's a lot of duplication of code because of the shared fields between all these rights basis. 

How about creating a rights basis instance, and then adding additional fields by randomly choosing from a list which contains tuples or dicts of the fields for each rights basis?

assign_rights/test_helpers.py
```diff
@@ -28,31 +34,103 @@ def add_groupings(count=5):
...
+        )
+
+    shell_types = [copyright_shell, other_shell, statute_shell, license_shell]
+    for x in range(count):
+        random.choice(shell_types)()
 
 
 def add_rights_acts(count=5):
     for x in range(count):
+        date_info = set_dates()
```
helrond: If you call this like:
```
start_date, end_date, start_date_period, end_date_period, end_date_open = set_dates()
```
you can then use those variables directly rather than using indexes. It makes it a bit more obvious what those variables are.

assign_rights/test_helpers.py
```diff
@@ -28,31 +34,103 @@ def add_groupings(count=5):
...
         RightsGranted.objects.create(
             basis=random.choice(RightsShell.objects.all()),
```
helrond, line 81: Do you want to make sure that *all* the RightsShells have rights granted associated with them or is it useful for that to be randomized? Which option reflects the anticipated state of data in the system?

helrond (changes requested):
Wondering about the nested function approach here, plus a few other minor things.

assign_rights/test_helpers.py
bonniegee: I want to make sure some RightsShells do _not_ have rights granted associated with them. Right now acts are created for only the first 5 (of 15) shells; does it make sense for this to be distributed differently?

helrond: If that's what you want to achieve, this test will achieve that. However, because you're selecting the RightShell you're attaching the RightsGranted to randomly, it's possible that you will end up less than 5 Rights Shells which have a RightsBasis, because you might grab the same one twice. Again, if that helps to mimic the shape of data you expect, that's fine.

assign_rights/test_helpers.py
```diff
@@ -58,79 +58,31 @@ def set_dates():
...
+        basis = random.choice(["copyright", "policy", "donor", "statute", "license"])
+        setattr(new_shell, "rights_basis", basis)
+        if basis in ["copyright", "statute"]:
+            setattr(new_shell, "jurisdiction", "us")
+        if basis == "copyright":
+            setattr(new_shell, "copyright_status", random.choice(["copyrighted", "public domain", "unknown"]))
+        if basis == "statute":
+            setattr(new_shell, "statute_citation", random_string())
+        if basis == "license":
+            setattr(new_shell, "license_terms", random_string())
```
helrond: Cool, this is a lot more terse, and I think makes more sense (at least to me).

A couple of things, just from a quick eye test:
- I would suggest instantiating (but not saving) the RightsShell in line 63. So, `new_shell = RightsShell(determination_date...)` and then calling `new_rights.save()` after you've added all the necessary fields. I'm not sure what you have here will actually persist those fields to the database.
- I'm not sure you need to use setattr - I think you can just do `new_shell.rights_basis = basis`, for example

assign_rights/test_helpers.py
```diff
@@ -60,17 +60,18 @@ def set_dates():
 def add_rights_shells(count=15):
     for x in range(count):
         start_date, end_date, start_date_period, end_date_period, end_date_open = set_dates()
-        new_shell = RightsShell.objects.create(determination_date=random_date(10, 0), note=random_string(), start_date=start_date, end_date=end_date, start_date_period=start_date_period, end_date_period=end_date_period, end_date_open=end_date_open)
+        new_shell = RightsShell(determination_date=random_date(10, 0), note=random_string(), start_date=start_date, end_date=end_date, start_date_period=start_date_period, end_date_period=end_date_period, end_date_open=end_date_open)
         basis = random.choice(["copyright", "policy", "donor", "statute", "license"])
         setattr(new_shell, "rights_basis", basis)
```
helrond:
```suggestion
        new_shell.rights_basis = basis
```

assign_rights/test_helpers.py
```diff
@@ -60,17 +60,18 @@ def set_dates():
...
+            new_shell.jurisdiction = "us"
         if basis == "copyright":
-            setattr(new_shell, "copyright_status", random.choice(["copyrighted", "public domain", "unknown"]))
+            new_shell.copyright_status = random.choice(["copyrighted", "public domain", "unknown"])
         if basis == "statute":
-            setattr(new_shell, "statute_citation", random_string())
+            new_shell.statute_citation = random_string()
         if basis == "license":
-            setattr(new_shell, "license_terms", random_string())
+            new_shell.license_terms = random_string()
```
helrond: does it make sense to make this `elif`s since they're mutually exclusive from line 68?

helrond (commented):
I'm okay with the contents of this PR. Did you want to merge in `issue-113` before merging this since tests are broken?

bonniegee:
@helrond I think that makes sense--I might point that PR against this branch so that the diff is clearer?

helrond (approved):

p-galligan (approved):

PR #116, opened 2021-03-19
Handle dates, not just terms in RightsAssembler

### LINKED ISSUE #113: RightsAssembler date methods need to handle dates, not just terms
The date methods (`calculate_dates` and `get_date_value`) assume that the start dates and end dates for a given basis or grant are `start_date_period` and `end_date_period`, and cannot currently handle other types of start and end dates.

Either before or in the `calculate_dates` method, check for start and end date types before calling `get_date_value` method.

bonniegee:
Changes to RightsAssembler so that multiple types of dates are added. Updates tests accordingly, including simplifying the check_object_dates method.

Fixes #113

### DIFF
--- a/assign_rights/assemble.py
+++ b/assign_rights/assemble.py
@@ -25,10 +25,10 @@ class RightsAssembler(object):
             shell_data = []
             for shell in rights_shells:
                 grant_data = []
-                start_date, end_date = self.calculate_dates(shell, request_start_date, request_end_date)
+                start_date, end_date = self.get_dates(shell, request_start_date, request_end_date)
                 serialized_shell = self.create_json(shell, RightsShellSerializer, start_date, end_date)
                 for grant in shell.rightsgranted_set.all():
-                    start_date, end_date = self.calculate_dates(grant, request_start_date, request_end_date)
+                    start_date, end_date = self.get_dates(grant, request_start_date, request_end_date)
                     grant_data.append(self.create_json(grant, RightsGrantedSerializer, start_date, end_date))
                 serialized_shell["rights_granted"] = grant_data
                 shell_data.append(serialized_shell)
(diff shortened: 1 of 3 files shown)


bonniegee:
Moved this back to draft after realizing tests are failing ~25% of the time. After adding a gazillion print statements, it looks like things go haywire when the start is a date, and the end is a term--working on that fix.

bonniegee:
Noting that this now mostly works, with the (notable) exception of when a term begins 0 years after creation date.

assign_rights/assemble.py
```diff
@@ -72,10 +55,16 @@ def calculate_dates(self, object, request_start_date, request_end_date):
                 representing the group of objects' start and end dates after
                 calculation.
         """
-        object_start = self.get_date_value(
-            object, "start_date", request_start_date, object.start_date_period)
-        object_end = None if object.end_date_open else self.get_date_value(
-            object, "end_date", request_end_date, object.end_date_period)
+        object_start = None
+        object_end = None
+        if getattr(object, "start_date_period"):
```
helrond: I would bet this is the offending line. Probably needs to be 
```suggestion
        if getattr(object, "start_date_period") is not None:
```

assign_rights/tests.py
```diff
@@ -193,37 +193,20 @@ def check_object_dates(self, object, request_start_date, request_end_date):
...
-        self.assertEqual(end_date, None)
-
-    def test_calculate_dates(self):
-        """Tests the calculate_dates method."""
+        if object.end_date_open:
+            self.assertEqual(end_date, None)
+        else:
+            self.assertTrue(isinstance(end_date, date))
+
+        if object.start_date_period:
```
helrond, line 203: Again, you may need a bit of a different test here:
```suggestion
        if object.start_date_period is not None:
```

assign_rights/assemble.py
bonniegee: I fixed this another way before seeing this--I flipped the order of which type of start is evaluated first.

helrond: OK, that's another way of solving that problem!

helrond (approved):

p-galligan (approved):

PR #137, opened 2021-04-12
Add default jurisdiction value of 'us'

### LINKED ISSUE #128: Add "us" as default value for jurisdiction
The vast majority of rights or statutes bases that we would create have a jurisdiction of "us". Setting this as a default value in the form would reduce the amount of work for a user.

For rights bases that include the jurisdiction field (copyright and statute), "us" should populate as the default value.

p-galligan:
Add default value to jurisdiction field. Fixes #128


helrond (approved):
Yup, I think this is the right approach, rather than specifying `default` in the model.

bonniegee (approved):
🌟

HaSistrunk (approved):

PR #138, opened 2021-04-12
Adds additional option to start date select

### LINKED ISSUE #127: Add option for starting immediately upon creation to start type drop-down
When creating a rights shell (or rights granted or restricted) that begins immediately upon creation date, it is unclear which start date type to select. 

Add an additional option to the start type drop down menu, to the effect of "Starts immediately upon end of creation date." When saved, this would correspond to a start term of zero years.

helrond:
Adds an option to the start date selects for both rights bases and acts which sets a `start_date_period` of zero years. While doing this, I also noticed that the detail pages don't display anything if the `start_date_period` value is zero, so I added some logic and custom display for that case.

fixes #127


p-galligan (approved):
Looks good to me.

HaSistrunk (approved):

HaSistrunk:
Cool, yeah this looks good.

bonniegee (approved):
+1 that this looks good!

PR #139, opened 2021-04-12
Better indication of rights IDs

### LINKED ISSUE #126: Rights IDs should be clearly indicated on grouping pages
On both the the index and individual grouping pages, it is hard to find the rights ID, even though it is included in the rights shell string method. Quickly finding a rights ID is part of the core functionality of the application, so this should be improved.

Clearly label the rights ID in all displays with text like "ID#". On the groupings index page, do not show the note text, as this information is not useful on that page.

One alternative is to remove the ID from the rights string totally, make the ID more prominent on the rights shell page, and link the rights shells from the grouping pages. The user would need to navigate to additional pages to get this information.

Another alternative is to remove the rights ID from the string method and add another column on the grouping table. This may be difficult to implement, since multiple rights shells are linked to a grouping.

A third alternative is to clearly label the rights ID in the string method, but not make any other changes.

HaSistrunk:
Fixes #126 

This adds "ID#" before all instances of the rights statement ID and the labeled identifier to the rights detail page. I didn't remove the note text on the Groupings page, because I'm wondering if consistency across the app is a plus, ie. every time the user sees this, the same information is clearly associated.

Take a look - I'm open to disagreement on this approach!

### DIFF
--- a/assign_rights/models.py
+++ b/assign_rights/models.py
@@ -50,7 +50,7 @@ class RightsShell(models.Model):
         note = self.note
         if len(note) > 115:
             note = "{}...".format(note[:75]) if len(note) > 75 else note
-        return "{} - {} ({})".format(self.pk, " / ".join([p for p in prefixes if p]), note)
+        return "ID# {} - {} ({})".format(self.pk, " / ".join([p for p in prefixes if p]), note)
 
 
 class RightsGranted(models.Model):
--- a/assign_rights/templates/rights/detail.html
+++ b/assign_rights/templates/rights/detail.html
@@ -12,6 +12,8 @@
 
 <h2>Details</h2>
 <dl class="row">
+  <dt class="col-sm-2">Identifier</dt>
+  <dd class="col-sm-10">{{ rightsshell.pk }}</dd>
   <dt class="col-sm-2">Basis</dt>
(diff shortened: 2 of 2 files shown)


helrond:
I'd also suggesting adding a (labeled) data element on the detail page.

More generally, I'm wondering if part of what's confusing is that we're presuming a number (an id) rather than a URI as an identifying piece of data. @bonniegee - I don't remember if Zorya just wants an ID, or if it needs the whole URI? If it's the former I am questioning our previous wisdom, because it seems like Zorya needs to know more about Aquila's API than is necessary? I'm not sure how this data is store/what other implications are though.

bonniegee:
@helrond As currently scoped, an archivist would add the IDs (envisioned as a number) to a bag-info.txt file. (Here's a [Zorya fixture](https://github.com/RockefellerArchiveCenter/zorya/blob/base/fixtures/bags/valid/digitization_bag.tar.gz) as an example.) I'm open to this changing a bit, but I think the core workflow of having an archivist take the IDs from Aquila and put them in the bag info file will be the same.

helrond (approved):

p-galligan (approved):

bonniegee (approved):

Issue #130, opened 2021-04-02
Difficult to select multiple rights shells from the grouping page


bonniegee:
It is difficult to select multiple rights shells on the grouping page. Because the field renders as a "select multiple" HTML attribute, a user must know how to select multiple options in their operating system (e.g., holding down the command key on a Mac).

There are 3 possible solutions:

- Adding help text that indicates that multiple options can be selected at once and directs the user how to select multiple options
- Changing the form widget from `SelectMultiple` to `CheckboxSelectMultiple`
- Customizing the `SelectMultiple` (or `CheckboxSelectMultiple`) widget

bonniegee:
Changing the form widget from `SelectMultiple` to `CheckboxSelectMultiple` is the preferred option. If that solution does not work, the other solutions should be evaluated.

PR #140, opened 2021-04-12
Improve selecting multiple rights shells when creating grouping

HaSistrunk:
Addresses issue #130 

Implements `CheckboxSelectMultiple` for the `rights_shells` in `forms.py`. However, by default this doesn't include the `<fieldset>` and `<legend>` elements that should form a set of checkboxes (this is a [useful reference](https://a11y-style-guide.com/style-guide/section-forms.html#kssref-forms-checkboxes) for that pattern, and why it's important), so I've edited the `templates/groupings/manage.html` template to add that markup.

Also enables vertical scroll for the checkboxes after a fixed height that encompasses 10 rights shells.

### DIFF
--- a/assign_rights/forms.py
+++ b/assign_rights/forms.py
@@ -1,7 +1,7 @@
 from crispy_forms.helper import FormHelper
-from crispy_forms.layout import Div, Field, Hidden, Layout
-from django.forms import (CheckboxSelectMultiple, ChoiceField, ModelForm, Select, Textarea, TextInput,
-                          inlineformset_factory)
+from crispy_forms.layout import Div, Field, Fieldset, Hidden, Layout
+from django.forms import (CheckboxSelectMultiple, ChoiceField, ModelForm,
+                          Select, Textarea, TextInput, inlineformset_factory)
 from django.forms.utils import ErrorList
 
 from .models import Grouping, RightsGranted, RightsShell
@@ -40,6 +40,14 @@ class RightsShellCommonLayout(Layout):
 
 
 class GroupingForm(ModelForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.helper = FormHelper(self)
+        self.helper.field_template = "forms/custom_field.html"
+        self.fields['rights_shells'].label = False
+        self.helper.form_tag = False
+        self.helper.disable_csrf = True
+
     class Meta:
         model = Grouping
         fields = ["title", "description", 'rights_shells']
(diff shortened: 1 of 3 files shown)


helrond:
@HaSistrunk easiest thing may be to tweak the existing `/templates/forms/custom_field.html` template.

helrond:
@HaSistrunk you can catch these linting errors before commit/push by doing `pre-commit install` locally, which will install the pre-commit hooks (which do the same linting as these checks in Travis CI).

HaSistrunk:
This is ready for a look, but I think it can be improved. I couldn't figure out how to cleanly implement the `<fieldset>` from the `/templates/forms/custom_field.html` template, but I've edited `/templates/groupings/manage.html` instead, removing the label for the rights_shells field from `forms.py`. This leaves the html a little messy with an extra `<div>`.

I've also added some Bootstrap styles that allow for scrolling to view a long set of rights shells (past 10), but the current max-height cut-off is coded in the html, which is probably not the ideal way to implement this?

p-galligan (commented):
I don't have any specific comments on the code placement/implementation.

However, I did notice that there's more padding underneath the last check item as opposed to above. I'm not sure if that's intended or not, but it looks slightly off to me. This may just be matter of preference though.

Screen Shot 2021-04-14 at 10 02 28 AM

assign_rights/forms.py
```diff
@@ -40,9 +40,17 @@ def __init__(self, *args, **kwargs):
 
 
 class GroupingForm(ModelForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.helper = FormHelper(self)
+        self.fields['rights_shells'].label = False
```
helrond: Is it possible to do this below in the `Meta` class?

HaSistrunk:
> I don't have any specific comments on the code placement/implementation.
> 
> However, I did notice that there's more padding underneath the last check item as opposed to above. I'm not sure if that's intended or not, but it looks slightly off to me. This may just be matter of preference though.


Yeah I see what you mean. This comes from the extra div (with `id=div_id_rights_shells`) that has styles with a margin-bottom value. Hmm, let me look at options.

assign_rights/templates/groupings/manage.html
```diff
@@ -5,8 +5,15 @@
 
 <form action="." method="POST" accept-charset="utf-8">
   {% csrf_token %}
-	{{ form | crispy }}
+	{{ form.title | as_crispy_field }}
+	{{ form.description | as_crispy_field }}
+	<fieldset>
+		<legend>Choose Associated Rights Shell/s*</legend>
+		<div class="overflow-auto border p-2 mb-3" style="max-height:250px;">
```
helrond: The way to avoid the inline styles here would be to either add the `max-height` to `.overflow-auto` or to add another class with that style.

HaSistrunk: Right. When I did that I couldn't make it work because of course it was a caching issue. Sigh.

helrond (approved):

p-galligan (approved):

bonniegee (approved):

PR #141, opened 2021-04-13
Serialize RightsShell jurisdiction to lowercase

### LINKED ISSUE #129: Country codes should be lowercased in the backend
When adding a country code to the jurisdiction field, a user can enter text of any cases. When serialized to JSON, this country code needs to be lowercased so that it conforms to the RAC rights schema.

In the rights serializer, the jurisdiction should be lowercased.

This could be enforced in the UI or when saving to the database.

This is related to #124 and #125

p-galligan:
Use the RightsShell serializer to convert jurisdiction codes to lowercase strings.

Adds logic to test the above to the RightsAssembler test.

Fixes #129

### DIFF
--- a/assign_rights/tests.py
+++ b/assign_rights/tests.py
@@ -228,9 +228,7 @@ class TestRightsAssembler(TestCase):
             serialized = self.assembler.create_json(obj, serializer_cls, start_date, end_date)
             if obj_cls == RightsShell:
                 if obj.jurisdiction:
-                    self.assertEqual(serialized['jurisdiction'], 'us')
-                else:
-                    self.assertEqual(serialized['jurisdiction'], None)
+                    self.assertTrue(serialized['jurisdiction'].islower())
             self.assertTrue(isinstance(serialized, dict))
             self.assertEqual(start_date, serialized["start_date"])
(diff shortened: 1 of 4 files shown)


assign_rights/tests.py
```diff
@@ -226,6 +226,11 @@ def test_create_json(self):
             start_date = random_date(75, 50).isoformat()
             end_date = random_date(49, 5).isoformat()
             serialized = self.assembler.create_json(obj, serializer_cls, start_date, end_date)
+            if obj_cls == RightsShell:
+                if obj.jurisdiction:
+                    self.assertEqual(serialized['jurisdiction'], 'us')
+                else:
+                    self.assertEqual(serialized['jurisdiction'], None)
```
helrond: I'd just assert that the string is lowercase if it exists. The second test clause isn't really testing anything ("if there's nothing there, assert there's nothing there"), and checking against a specific string is probably a little too restrictive here.
```suggestion
                if obj.jurisdiction:
                    self.assertTrue(serialized['jurisdiction'].islower())
```

helrond (approved):

bonniegee (approved):

HaSistrunk (approved):

PR #144, opened 2021-04-14
Create and call basis-specific serializers

bonniegee:
Add basis-specific serializers that inherit from RightsShellSerializer. I'd be interested in feedback on making this less verbose. Since this does not address policy or donor bases yet, the tests bypass validating rights shells of those types.

Fixes #125, addresses #47

PR #147, opened 2021-04-15
Add serializer to handle policy and donor bases

### LINKED ISSUE #47: Validate rights JSON in tests

### LINKED ISSUE #124: Rights serializer is not creating expected output for policy, donor basis
The rights serializer is currently creating JSON with "policy" and "donor" as a rights basis, instead of a rights basis of "other" with "policy" or "donor" as the other basis type. This is because "policy" and "donor" are stored as their own rights basis in the rights shell model, so some additional transformational work needs to be done before the JSON is properly serialized.

The rights serializer (or some method elsewhere?) should handle "policy" and "donor" rights shells by serializing them to a rights basis of "other" with "policy" or "donor" as the other basis type.

bonniegee:
Adds "other" serializer which returns a rights basis of "other" and an other basis of "policy" or "donor."

Fixes #124. Fixes #47.

Note: should be merged after #144.

### DIFF
--- a/assign_rights/serializers.py
+++ b/assign_rights/serializers.py
@@ -56,11 +56,10 @@ class RightsShellSerializer(serializers.ModelSerializer):
 
 class CopyrightSerializer(RightsShellSerializer):
     jurisdiction = serializers.SerializerMethodField()
-    status = serializers.CharField(source="copyright_status")
 
     class Meta:
         model = RightsShell
-        fields = RightsShellSerializer.Meta.fields + ('determination_date', 'jurisdiction', 'status')
+        fields = RightsShellSerializer.Meta.fields + ('determination_date', 'jurisdiction', 'copyright_status')
 
     def get_jurisdiction(self, obj):
         return obj.jurisdiction.lower() if obj.jurisdiction else None
@@ -78,8 +77,7 @@ class StatuteSerializer(RightsShellSerializer):
 
 
 class LicenseSerializer(RightsShellSerializer):
-    terms = serializers.CharField(source="license_terms")
 
     class Meta:
         model = RightsShell
-        fields = RightsShellSerializer.Meta.fields + ('terms',)
+        fields = RightsShellSerializer.Meta.fields + ('license_terms',)
--- a/assign_rights/serializers.py
+++ b/assign_rights/serializers.py
@@ -28,25 +28,58 @@ class RightsShellSerializer(serializers.ModelSerializer):
     """
     start_date = serializers.CharField()
     end_date = serializers.CharField()
-    status = serializers.CharField(source="copyright_status")
-    terms = serializers.CharField(source="license_terms")
+    # terms = serializers.CharField(source="license_terms")
     rights_granted = serializers.ListField(default=[])
-    jurisdiction = serializers.SerializerMethodField()
 
     class Meta:
         model = RightsShell
         fields = (
             "rights_basis",
-            "determination_date",
-            "jurisdiction",
             "start_date",
             "end_date",
             "note",
-            "status",
-            "terms",
-            "statute_citation",
             "rights_granted"
         )
+        # fields = (
+        #     "rights_basis",
+        #     "determination_date",
+        #     "jurisdiction",
+        #     "start_date",
+        #     "end_date",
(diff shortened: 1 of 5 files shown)


p-galligan (approved):

assign_rights/serializers.py
```diff
@@ -28,26 +28,58 @@ class RightsShellSerializer(serializers.ModelSerializer):
...
             "note",
-            "status",
-            "terms",
-            "citation",
             "rights_granted"
         )
 
+
+class OtherSerializer(RightsShellSerializer):
+    rights_basis = serializers.SerializerMethodField()
```
helrond, line 49: I don't actually know if it's possible, but can you just set this to a value?
```suggestion
    rights_basis = "other"
```

assign_rights/serializers.py
```diff
@@ -28,26 +28,58 @@ class RightsShellSerializer(serializers.ModelSerializer):
...
+    other_basis = serializers.SerializerMethodField()
```
helrond: You can do this much more simply: 
```suggestion
    other_basis = serializers.CharField(source='rights_basis')
```

https://www.django-rest-framework.org/api-guide/serializers/#specifying-fields-explicitly

assign_rights/serializers.py
```diff
@@ -28,26 +28,58 @@ class RightsShellSerializer(serializers.ModelSerializer):
...
+
+class StatuteSerializer(RightsShellSerializer):
+    jurisdiction = serializers.SerializerMethodField()
+
+    class Meta:
+        model = RightsShell
+        fields = RightsShellSerializer.Meta.fields + ('determination_date', 'jurisdiction', 'statute_citation')
+
+    def get_jurisdiction(self, obj):
+        return obj.jurisdiction.lower() if obj.jurisdiction else None
```
helrond: Since this shows up in multiple classes that are based on RightsShellSerializer I think it should be possible to define this method on RightsShellSerializer and have it inherit down. Worth a try.

helrond (changes requested):
This will be easier to review once #144 is merged in - right now it's a a little hard to distinguish the changes from one another. I've suggested a few tweaks and somethings that _might_ not work but are worth trying.

assign_rights/serializers.py
bonniegee: Unfortunately, no, it cannot be set directly to a value.

helrond: Can you do `rights_basis = serializers.CharField(default="other")` ?

bonniegee: That doesn't work either.

assign_rights/tests.py
```diff
@@ -219,14 +219,18 @@ def test_get_dates(self):
 
     def test_create_json(self):
         """Tests that Serialzers are working as expected."""
-        for obj_cls, serializer_cls in [
-                (RightsShell, RightsShellSerializer),
-                (RightsGranted, RightsGrantedSerializer)]:
+        for obj_cls in [RightsShell, RightsGranted]:
             obj = random.choice(obj_cls.objects.all())
```
helrond: Just noticing this - since the RightsShells now have different serializers based on the rights_basis, I think we'll want  to make sure we test with a RightsShell from each of the four rights bases.

assign_rights/serializers.py
helrond: Ah okay, looking at the DRF docs a little more it does seem like a `SerializerMethodField` is the recommended approach.

assign_rights/tests.py
bonniegee: I'm looking at this now, and am wondering if we need to do a different setup for adding rights shells to make sure that we get at least one of each basis, or add additional shells as part of this method. Does that sound right?

helrond: Yes I think so. I would think doing it in the setup probably makes more sense, as that gets used across multiple tests and will make everything better?

helrond (approved):
Looking good!

HaSistrunk (approved):

PR #149, opened 2021-04-16
Adds API templates

### LINKED ISSUE #148: Add templates for API
The assemble-rights endpoint does not have a template associated with it, and is throwing a `TemplateDoesNotExist` error. 

Add the appropriate template(s) so the API is browsable.

helrond:
Adds templates for DRF, fixes #148 
Adds `rest_framework` to installed apps, so that tags can be loaded.
Updates existing nav so it can be used across site.


bonniegee:
This looks good! A couple of things from the UI:

- I'm wondering if the name for this page in the navigation bar should be something like "API" instead of "Rights Assembler," so its purpose is clearer to a user. (This can also just be something we look for in user testing.)
- I'm not sure if this is even an issue (or just a documentation thing), but I'm getting a `"detail": "CSRF Failed: CSRF token missing or incorrect."` error when I attempt to post JSON.

helrond:
OK, I did the following:
- Changed "Rights Assembler" to "Rights Assembler API" as I think it's a bit misleading to call this an API if it only has one endpoint (we also link directly to just that endpoint rather than a router home page).
- Disabled session authentication in DRF, which is what was causing that CSRF error.

p-galligan (approved):

assign_rights/templates/rest_framework/api.html
```diff
@@ -0,0 +1,241 @@
...
+      </nav>
+    {% endif %}
+
+    <div class="request-info" style="clear: both" aria-label="{% trans "request info" %}">
+      <pre class="prettyprint"><b>{{ request.method }}</b> {{ request.get_full_path }}</pre>
+    </div>
+
+    <div class="response-info" aria-label="{% trans "response info" %}">
+      <pre class="prettyprint"><span class="meta nocode"><b>HTTP {{ response.status_code }} {{ response.status_text }}</b>{% autoescape off %}{% for key, val in response_headers|items %}
+<b>{{ key }}:</b> <span class="lit">{{ val|break_long_headers|urlize }}</span>{% endfor %}
```
HaSistrunk: Not to nitpick, but how about using `<span class="font-weight-bold">` instead of `<b>` to make these bold for style instead of semantics?

HaSistrunk:
Great. I just made one small suggestion above, and it looks like it needs the skip link added to the template.

bonniegee (approved):

assign_rights/templates/rest_framework/api.html
helrond: I've removed this block from the template and we're just using the base code provided by DRF for this block now.

HaSistrunk (approved):

PR #155, opened 2021-05-05
Overrides save method to lowercase jurisdiction

### LINKED ISSUE #151: Lower case jurisdiction should be enforced in model, not serializers
Upon further discussion of #129, country codes should be enforced at the database level, and not at the serializer level.

Update the model so that country codes are saved and stored as lower case.

The serializer method can be removed.

helrond:
Implements a custom save method which lowercases the `jurisdiction` value if it exists. Removes the serializer code which coerced values to lowercase.

fixes #151

### DIFF
--- a/assign_rights/models.py
+++ b/assign_rights/models.py
@@ -39,6 +39,11 @@ class RightsShell(models.Model):
     def get_absolute_url(self):
         return reverse("rights-detail", kwargs={"pk": self.pk})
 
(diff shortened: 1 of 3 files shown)


p-galligan (approved):
Looks good to me. I checked out the object values in the shell and it worked.

HaSistrunk (approved):

bonniegee (approved):

PR #161, opened 2021-06-03
Improve display of "browse" and "create" buttons on cards on homepage

bonniegee:
Removes footer on cards and makes buttons full width (and stack vertically). I do not like that these buttons do not align across cards, but any attempt to fix caused more problems.

### DIFF
--- a/assign_rights/templates/index.html
+++ b/assign_rights/templates/index.html
@@ -6,24 +6,33 @@
 <div class="row mb-3">
   <div class="col">
     <div class="card full-height">
-      <div class="card-body d-flex flex-column">
-        <h2>Groupings</h2>
-        <p>
-          Aggregations of records with the same rights statuses. Groupings don't
-          necessarily equate to record groups, and are not required to be from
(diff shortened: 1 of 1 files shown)


helrond:
OK, here's what I did:
- wrap the text and buttons each in a div.
- having those two divs then allowed me to use the flexbox `justify-content: space-between` to push the text to the top and the buttons to the bottom. I did this by using a Bootstrap [flex utility](https://getbootstrap.com/docs/4.1/utilities/flex/), meaning I added an additional class (`justify-content-between`) because that's how Bootstrap rolls.

Flexbox is incredibly powerful and incredibly confusing. I have found the [CSS Tricks](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) very useful, and find myself going back to it often.

bonniegee:
Thank you! It looks like exactly what I was trying to get; as you mentioned, I was getting a bit stuck with the flex utility.

helrond (approved):

PR #164, opened 2021-06-08
Render linebreaks in HTML

### LINKED ISSUE #163: Line breaks in grouping description are not visible on view pages
When a description for a grouping with line breaks is saved, the line breaks are not visible on the grouping index page nor on the individual grouping page. They are still visible when editing the record.

Screen Shot 2021-06-07 at 13 54 27
Screen Shot 2021-06-07 at 13 54 41
Screen Shot 2021-06-07 at 13 55 11

I would like line breaks to display on the index and individual record view pages for ease of reading. I'm not sure if there's an easy way to do this, and where the fix would be.

helrond:
Uses the Django template tag [linebreaks](https://docs.djangoproject.com/en/3.1/ref/templates/builtins/#linebreaks) to add HTML line breaks.

Fixes #163

### DIFF
--- a/assign_rights/templates/groupings/detail.html
+++ b/assign_rights/templates/groupings/detail.html
@@ -6,7 +6,7 @@
 <small>Created on {{ object.created }}</small><br/>
 <small>Last modified on {{ object.last_modified }}</small>
 
-<p>{{ object.description }}</p>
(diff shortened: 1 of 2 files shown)


bonniegee (approved):
Looks great!

PR #165, opened 2021-06-08
Underline navigation links on hover

### LINKED ISSUE #157: Items in navigation bar are difficult to distinguish from one another
The three navigation items in the navigation bar are difficult to distinguish from one another. This is due to the similarity of the text color when a user hovers over an item, and possibly due to the spacing of the items.

Change the color of the text that is hovered over, so it stands out more. May need to also increase spacing between items.

helrond:
Underlines navigation links on hover. This seemed to be the cleanest intervention here - changing the color of the text on hover while still meeting accessibility baselines for color contrast was challenging, and the spacing between nav items is tied into responsiveness of the navbar as a whole, so I took the old-school approach here.

fixes #157

### DIFF
--- a/assign_rights/static/css/custom.css
+++ b/assign_rights/static/css/custom.css
@@ -70,3 +70,7 @@ a.btn {
 .full-height {
   height: 100%;
 }
+
+.nav-link:hover {
+  text-decoration: underline;
+}


HaSistrunk (approved):
This makes sense to me. I think the underline creates a clear visual queue that distinguishes the nav links once the user moves to click on one of them, which should help address the problem we observed in testing. And I see the keyboard focus styles are the browser defaults, which work fine here.

bonniegee (approved):
I think this looks good!

PR #167, opened 2021-06-08
Allow default user to modify other users' permissions

### LINKED ISSUE #158: Set up user who can modify users' permissions
There is no way for one user to modify the permissions of other users. Right now, you need to manually create a [super user](https://developer.mozilla.org/en-US/docs/Learn/Server-side/Django/Admin_site#creating_a_superuser) and log in as that user in order to give any other account permission to see the Django admin site and modify other users permissions.

When running in production, we'll use single sign-on (see #37), so while we will not need to manually create new users, we will need to modify their permissions.

On setup, there needs to be an initial account created that has the `assign_rights | user` permissions enabled.

## Additional context
Given the small number of users of this application, and that any additional users will be rarely added at all (let alone need additional permissions), using the Django admin site is sufficient for managing user permissions.

helrond:
Allows the initial user to update other user's groups and permissions. Because we are going to be doing this through the Django admin, the user also needed to have `is_staff` permissions set, so it seemed like a cleaner way to do this was just to create a superuser rather than trying to sort out all the permissions and groups things.

I've also made the superuser name and email a config, added a note to the README to advise changing the password for the default superuser, and added a link to the Django admin interface for users who are staff.

fixes #158


bonniegee (approved):
This all looks great!

PR #171, opened 2021-06-23
Hides create buttons on home page from users without write privileges

### LINKED ISSUE #170: Users with view-only permissions can see (and click on) "create" buttons on home page
## Describe the bug
If a user does not have permissions to "create" a new rights shell or grouping, they are still able to see the "create" buttons on the homepage. When that user clicks on one of the buttons, they are taken to a 403 error page.

## To reproduce
1. Login as a user who is not part of the "edit" group
2. Go to the home page
3. Click on the "Create a new grouping" or "Create a new rights statement" page

## Expected behavior
Users who cannot "create" new records should not be able to see the related buttons (or the related buttons can be disabled).

## Error message or screenshots
Screen Shot 2021-06-22 at 18 37 28

helrond:
Hides the "create" buttons on the home page if users do not have edit permissions.

Also updates the `has_group` template tag so that superusers can see everything, as would be expected, and adds the `DEFAULT_AUTO_FIELD` value to the config.

Fixes #170

### DIFF
--- a/aquila/settings.py
+++ b/aquila/settings.py
@@ -159,3 +159,5 @@ LOGIN_REDIRECT_URL = "home"
 
 # django-crispy-forms template pack
 CRISPY_TEMPLATE_PACK = 'bootstrap4'
+
+DEFAULT_AUTO_FIELD = 'django.db.models.AutoField'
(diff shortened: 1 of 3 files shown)


p-galligan (approved):
Looks good to me. I hadn't thought about the superuser angle.

bonniegee (approved):
Looks great!

PR #186, opened 2021-09-02
Add GROUP_CLAIMS setting

helrond:
This should solve the issue where user groups are reset. It's hard to fully test without deploying though.


bonniegee (approved):
This looks good--thank you!!

PR #191, opened 2021-09-16
Update Travis config

bonniegee:
Updates Travis config as part of CI maintenance. I was getting a bit tripped up by using `PORT` as a variable, since that variable is also used in deployment. Do I not need to worry about it since it will be overwritten?

The README will need to be updated, but I'll do that in a separate PR as I want to move this forward before finishing setting up deployment on production (and making these changes in pictor).

I know there is an open dependency PR but I decided not to include that fix here, since the most recent version of Django is still using sqlparse `0.4.1`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
   - echo "${DOCKER_PASSWORD}" | docker login -u "${DOCKER_USERNAME}" --password-stdin
   - docker-compose up -d
 install:
-  - ./wait-for-it.sh ${SERVICE_NAME}:8009 -- docker-compose exec ${SERVICE_NAME} pip install coverage
+  - ./wait-for-it.sh ${SERVICE_NAME}:8009 -- docker-compose exec ${SERVICE_NAME} pip install coverage pre-commit
   - docker-compose exec ${SERVICE_NAME} pre-commit install
 script:
   - docker-compose exec ${SERVICE_NAME} pre-commit run --all-files --show-diff-on-failure
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ script:
   - docker-compose exec ${SERVICE_NAME} pre-commit run --all-files --show-diff-on-failure
   - docker-compose exec ${SERVICE_NAME} coverage run manage.py test
 after_script:
-  - docker-compose exec $CONTAINER coverage report --omit=*/migrations/* -m
+  - docker-compose exec ${SERVICE_NAME} coverage report --omit=*/migrations/* -m
 before_deploy:
   - sudo deploy_scripts/substitute_env.sh
   - sudo deploy_scripts/make_zip.sh $DEPLOY_ZIP_DIR $DEPLOY_ZIP_NAME
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ deploy:
     access_key_id: $AWS_ACCESS_KEY
     secret_access_key: $AWS_SECRET_KEY
     application: ${APPLICATION_NAME}
-    deployment_group: AquilaDevelopment
+    deployment_group: $DEPLOYMENT_GROUP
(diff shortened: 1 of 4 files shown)


helrond (changes requested):
Cool this looks good! I like the addition of the APPLICATION_NAME variable, and am going to roll that into my changes to Zodiac.

Regarding the port - the environment variables you set in Travis CI are available here, so you could just target $DJANGO_PORT. Again, I'll update that in Zodiac.

Can you add an `after_script:` section to print out the coverage report? That helps us measure how good unit test coverage is. See [zodiac](https://github.com/RockefellerArchiveCenter/zodiac/blob/ci/.travis.yml#L18-L19)

Last, it is possible to update sqlparse - you'll need to delete the requirements.txt file and run `pip-compile`.

bonniegee:
> Regarding the port - the environment variables you set in Travis CI are available here, so you could just target $DJANGO_PORT. Again, I'll update that in Zodiac.

In the case of $DJANGO_PORT, we have that value as specific to each branch for this application, which would mean we would need to set it in this file for Travis to successfully run on other branches?

helrond:
> In the case of $DJANGO_PORT, we have that value as specific to each branch for this application, which would mean we would need to set it in this file for Travis to successfully run on other branches?

Yes, that's right - good point. I wonder if there's a way for us to more reliably tie all of the ports together - there's stuff in configs but also the entrypoint uses a port number that is hardcoded. And maybe other places as well?

docker-compose.yml
```diff
@@ -11,7 +11,7 @@ services:
     volumes:
       - .:/code
     ports:
-      - "8000:8000"
+      - "8009:8000"
```
helrond:
```suggestion
      - "8009:8009"
```

.travis.yml
```diff
@@ -1,27 +1,23 @@
 language: python
+cache: pip
 services:
   - docker
 env:
   global:
-    - DOCKER_COMPOSE_VERSION: 1.25.5
     - APPLICATION_NAME: aquila
     - SERVICE_NAME: aquila-web
```
helrond: Two things: 
- we've called the `SERVICE_NAME` variable `CONTAINER` in other contexts and I'd prefer to be consistent. 
- I think it makes sense to use a `DJANGO_PORT` variable here, even though there's also a environment variable already defined in Travis for deployment. Getting these ports all aligned as variables is going to take some more thinking, but in the meantime it's useful to have that pulled out as a variable so we remember we have to deal with it.

.travis.yml
```diff
@@ -1,27 +1,23 @@
...
-  - sudo mv docker-compose /usr/local/bin
   - cp ${APPLICATION_NAME}/config.py.example ${APPLICATION_NAME}/config.py
-  - echo "$DOCKER_PASSWORD" | docker login -u "$DOCKER_USERNAME" --password-stdin
-  - docker-compose build
+  - echo "${DOCKER_PASSWORD}" | docker login -u "${DOCKER_USERNAME}" --password-stdin
   - docker-compose up -d
-  - docker-compose exec ${SERVICE_NAME} pip install pre-commit
-  - docker-compose exec ${SERVICE_NAME} pip install coverage
+install:
+  - ./wait-for-it.sh ${SERVICE_NAME}:8009 -- docker-compose exec ${SERVICE_NAME} pip install coverage pre-commit
```
helrond: Use a variable for the port number as described above.

.travis.yml
```diff
@@ -33,7 +29,8 @@ deploy:
     skip_cleanup: true
     on: &development
```
helrond:
```suggestion
    on:
```

helrond (changes requested):
A couple of minor things.

helrond (approved):

HaSistrunk (approved):

p-galligan (approved):

PR #204, opened 2021-11-12
Prep for Google Analytics

### LINKED ISSUE #175: Set up web analytics
Since this has a UI and we've done usability testing, should we set up an analytics property and add a tracking id to gather analytics data?

bonniegee:
Fixes #175 

Since this is deployed continuously, we'll want to set up the GTM ID in Travis before merging. @HaSistrunk is it correct that we can use a dummy ID for dev? And where can I find the GTM ID for prod? I just did a quick spin through the docs site and Teams and couldn't find anything.

### DIFF
--- a/aquila/config.py.deploy
+++ b/aquila/config.py.deploy
@@ -13,3 +13,5 @@ SSO_AUDIENCE = "${SSO_AUDIENCE}"
 LOGIN_URL = "django_auth_adfs:login"
 SUPERUSER_USERNAME = "${SUPERUSER_USERNAME}"
 SUPERUSER_EMAIL = "${SUPERUSER_EMAIL}"
+# Google analytics configs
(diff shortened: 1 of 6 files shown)


HaSistrunk:
> @HaSistrunk is it correct that we can use a dummy ID for dev?

The analytics property is set up in Google Analytics to filter out non-production traffic data by only including data from the hostname pattern `^aquila\.rockarch\.org$`, so we don't have to use a dummy id for dev. Including the correct ID in dev enables us test the analytics setup in dev before merging to prod.

> And where can I find the GTM ID for prod?

It's the container id for the aquila property in Google Tag Manager (tagmanager.google.com): 
`GTM-KV7B9FH`

helrond (approved):

HaSistrunk (approved):

PR #303, opened 2024-03-22
Implement RAC Style Library

### LINKED ISSUE #298: Implement Style Library
Where possible, replace local styles with the [RAC Style Library](https://github.com/RockefellerArchiveCenter/styles) styles: [https://assets.rockarch.org/v0.9.1/main.min.css.](https://assets.rockarch.org/v0.8.1/main.min.css). 

Update UI components to follow the [RAC Style Guide
](https://styles.rockarch.org/?path=/docs/components-alert--docs)

HaSistrunk:
Fixes #298 

Implements the RAC Style Library v0.10.0.

- Removes Bootstrap styles and associated scripts
- Adds custom script for toggling mobile nav
- The manage rights and groupings forms are still using the bootstrap 4 crispy template pack because the layout function in forms.py is very useful here, but I've adjusted the custom_field.html to bring in some Style Library classes, and adjusted the form error message css from bootstrap to match ours.


HaSistrunk:
Note: I did choose to use color variables in my local styles. I was thinking this would make maintenance easier if there were updates to the Style Library so that someone could quickly see what a color was in our Style Guide by name. Happy to be talked out of that, though.

helrond (commented):
I haven't gotten super into the weeds on the accessibility side of things, but from a general HTML/Django perspective this all makes sense to me (and this looks quite slick!). 

A couple of thoughts:
- It looks like you grabbed `rac_base.html` and `api.html` from the Argo repo, which makes sense. HOWEVER a couple of things: the endpoint that is available in the UI is a POST route, but there is no POST form, so it's basically a useless page. I'm divided on whether this page should just go away, or if it makes sense to keep it around but add that form back in...I'm not sure what the thinking was on having it in the UI in the first place but at this point the utility seems limited.
- If you _do_ end up keeping the API UI, we should remove the div with the Request Parameters header because that is specific to the RAC API and doesn't provide much info here. I don't hate the headers for the page sections as much as I thought I would, so definitely something to evaluate for Argo as well.

helrond:
Oh, and let me know if you have issues with figuring out how to associate the inputs and error messages. The data should be available to you in the templates.

HaSistrunk:
> I'm divided on whether this page should just go away, or if it makes sense to keep it around but add that form back in...I'm not sure what the thinking was on having it in the UI in the first place but at this point the utility seems limited.

Looking back at the issues and PRs in this repo re: the API, there isn't a lot to go on, but it seemed like we just wanted to make the [endpoints browseable](https://github.com/RockefellerArchiveCenter/aquila/issues/148). You more recently added an [API view for rights shells](https://github.com/RockefellerArchiveCenter/aquila/pull/284/files). Now that aquila is up and running, if we aren't using the Rights Assembler API I'm happy to get rid of it, but I don't think I should be the one to make that call.

HaSistrunk:
> Oh, and let me know if you have issues with figuring out how to associate the inputs and error messages. The data should be available to you in the templates.

Ok, so I tried to edit templates to do this, but I'm struggling to find a way to make it work. Basically, the error message divs already have associated ids, so I was trying to add an `aria-describedby` to the form input that is the same as that id. I feel like there might be an easier way to do it in views.py, but I can't figure it out. I also looked at using [crispy-boostrap5](https://github.com/django-crispy-forms/crispy-bootstrap5) instead, but there were no improvements in version 5 for form validation accessibility. [Django 5 core has made progress on this](https://www.djangoproject.com/weblog/2024/feb/10/django-accessibility-in-2023-and-beyond/), though.

Any help would be much appreciated!

HaSistrunk:
Upon further consideration, it may make sense to get rid of crispy forms altogether. There's a lot to extract and replace from crispy forms here, though.

HaSistrunk:
Update: I've removed crispy forms, but am struggling to adjust the RightsShellUpdateForm to render correctly with the rights start and end date fields.

HaSistrunk:
Final changes here and this is ready for review:
- I removed the browseable API because the team determined it wasn't used or useful.
- I decided to keep crispy forms, because the Layout feature that we are using in forms.py to create a Right Shell Common Layout is incredibly useful. Our form error message patterns may need some accessibility work in the future, but there are some variations in approach that may be useful to look at together across sites and in the Style Guide.

p-galligan (approved):

assign_rights/forms.py
```diff
@@ -28,58 +30,45 @@ class Meta:
...
-    dates = RightsShellDates()
+                 ("end_date_open", "There is no end date for these rights")))
 
     def __init__(self, *args, **kwargs):
-        kwargs.setdefault('label_suffix', '')
         super().__init__(*args, **kwargs)
+        self.helper = FormHelper(self)
+        self.helper.field_template = "forms/custom_field.html"
+        self.helper.form_tag = False
+        self.helper.disable_csrf = True
```
helrond: This seems...unexpected. Any insight into why this is disabled?

helrond (changes requested):
Just one question (which, if there's a good reason for, should just be documented as a code comment).

assign_rights/forms.py
HaSistrunk: It was added in PR #92 , but no specific explanation there or in other PRs/issues that I can find. And I see in the template that we have added {% csrf_token %} in the [rights/manage.html form](https://github.com/RockefellerArchiveCenter/aquila/blob/55f2311267346732badfe51a5bdd0c36c12737a4/assign_rights/templates/rights/manage.html#L9), so that conflicts (but the disable csrf is taking precedence here with the crispy form I think). I can't figure out why we would want that, so my inclination is to remove this line in forms.py unless you can recall a reason?

helrond: I can't recall a reason - I guess just make sure the form is still functional after you remove it?

helrond (approved):

PR #383, opened 2026-04-08
Fixing rights detail layout issues

helrond:
Corrects HTML on rights detail page so that rights statements with multiple acts now display as expected. Screenshot below.
image


HaSistrunk:
This looks like a straightforward fix to me, but I'm struggling to run the UI in the container locally to test. I think there's something off with the ports?

HaSistrunk (approved):

